PR #6, opened 2021-08-17
Allow configuration of some parameters via config object

jefferai:
Right now a few parameters can only be configured via env vars. This
allows configuration to be set via the config object as well.


austingebauer (approved):
👍

PR #7, opened 2021-09-02
Add rotation and a few related helper functions

jefferai:
This was pulled out of Vault's codebase, then enhanced to split
different parts of it into different functions. This also introduces the
option pattern. The rotation test validates creation and deletion
independently as well as the combination via the rotate function.

Coverage (when running with real AWS creds) is mostly green for all
added code, except for places where we are handling exceptional
conditions from AWS' API.


vancluever (approved):

awsutil/options.go
```diff
@@ -0,0 +1,161 @@
...
+	withStsEndpoint            string
+	withIamEndpoint            string
+	withMaxRetries             *int
+	withRegion                 string
+	withHttpClient             *http.Client
+}
+
+func getDefaultOptions() options {
+	return options{
+		withEnvironmentCredentials: true,
```
calvn, line 47: Do functional options tend to strictly stick with `With*`, or can we swap these to something like `SkipEnvironmentCredentials` so that the default value is also the zero value (false)?

Something like this seems to be a bit better to read, especially if this is for opt-out logic.
```
SkipEnvironmentCredentials(true)
```

jefferai: Default value isn't related to `With`. The function takes in a boolean, so it's not `WithEnvironmentCredentials` always results in true, it results in whatever you pass it.

awsutil/generate_credentials.go
```diff
@@ -55,18 +62,65 @@ type CredentialsConfig struct {
...
 
+	// The max retries to set on the client. This is a pointer because the zero
+	// value has meaning. A nil pointer will use the default value.
+	MaxRetries *int
+
 	// The logger to use for credential acquisition debugging
 	Logger hclog.Logger
 }
 
+func NewCredentialsConfig(opt ...Option) (*CredentialsConfig, error) {
```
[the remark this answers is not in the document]
jefferai: Ah yeah, will do.

vancluever (dismissed):

awsutil/options.go
calvn: Right, I realize that `getDefaultOptions` will set these to `true` if that's not provided and will honor what's being provided otherwise. I just thought that it'd be nicer to have something like this

```
SkipEnvironmentCredentials(true)
or
SkipEnvironmentCredentials()
```

than

```
WithEnvironmentCredentials(false)
```
since this behaves as an opt-out. This is all subjective though, so I'm fine with leaving as-is.

awsutil/generate_credentials.go
```diff
@@ -32,6 +33,12 @@ type CredentialsConfig struct {
 	// The session token if it is being used
 	SessionToken string
 
+	// The IAM endpoint to use; if not set will use the default
+	IamEndpoint string
```
calvn, line 37: Nit: Should we capitalize to `IAMEndpoint` (same with `STSEndpoint`) in the same spirit that we have for `HTTPClient`?

calvn (approved):
Few small comments, otherwise looks good!

awsutil/options.go
jefferai: I got convinced by some other people that functions toggling booleans should always take in boolean parameters instead of no parameters, since you may want to explicitly toggle something to the same as the default (e.g. for clarity, or if you don't know what the default is). Hence sticking with `With`. As you said it's subjective, so I'm going to leave it this way, but I get where you're coming from :-)

awsutil/generate_credentials.go
jefferai: Yeah I guess I can do that. I wish I could change `HTTPClient` instead, but alas :-)

PR #9, opened 2021-09-16
awsutil: add GetCallerIdentity

vancluever:
This adds the GetCallerIdentity call, which allows for a set of
credentials/session to get information on its effective account ID and
principal ARN.

This can be useful in checking the validity of credentials, as it
requires no permissions to run other than the access key to be usable.

For info on the call, see:
  https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html

### DIFF
--- a/awsutil/options.go
+++ b/awsutil/options.go
@@ -3,6 +3,7 @@ package awsutil
 import (
 	"fmt"
 	"net/http"
+	"time"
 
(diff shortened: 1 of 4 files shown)


jefferai (approved):
Looks good!

vancluever:
@jefferai looks like I can't merge to this repository - can I get permission or can you do it?

Thanks!

PR #10, opened 2021-09-17
awsutil: add GetCallerIdentity to Create/Rotate

vancluever:
This adds GetCallerIdentity to CreateAccessKey and RotateKeys. The
latter is done simply by passing WithTimeout through to the inner
CreateAccessKey on Rotate. As such, both paths are conditional on a
non-zero WithTimeout option being given - this is to ensure that we
don't try to immediately verify, which is likely to always fail.

### DIFF
--- a/awsutil/rotate.go
+++ b/awsutil/rotate.go
@@ -23,8 +23,8 @@ import (
 // if the old one could not be deleted.
 //
 // Supported options: WithEnvironmentCredentials, WithSharedCredentials,
-// WithAwsSession, WithUsername, WithTimeout. Note that WithTimeout
-// here, when non-zero, controls the WithTimeout option on access key
+// WithAwsSession, WithUsername, WithValidityCheckTimeout. Note that WithValidityCheckTimeout
+// here, when non-zero, controls the WithValidityCheckTimeout option on access key
 // creation. See CreateAccessKey for more details.
 func (c *CredentialsConfig) RotateKeys(opt ...Option) error {
 	if c.AccessKey == "" || c.SecretKey == "" {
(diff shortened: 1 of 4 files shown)


awsutil/rotate.go
```diff
@@ -114,6 +114,20 @@ func (c *CredentialsConfig) CreateAccessKey(opt ...Option) (*iam.CreateAccessKey
...
+	// Check the credentials to make sure they are usable. We only do
+	// this if withTimeout is non-zero to ensue that we don't
+	// immediately fail due to eventual consistency.
+	if opts.withTimeout != 0 {
+		newC := &CredentialsConfig{
+			AccessKey: *createAccessKeyRes.AccessKey.AccessKeyId,
+			SecretKey: *createAccessKeyRes.AccessKey.SecretAccessKey,
+		}
+
+		if _, err := newC.GetCallerIdentity(WithTimeout(opts.withTimeout)); err != nil {
```
jefferai, line 126: So thinking about this, the one thing I'm worried about is the exponential backoff. The reason is that this often takes 3-5 seconds but you could easily end up sleeping 4-8 seconds more than you need to. I wonder if we should just fix the retry period at something shorter (maybe 1 or 2 seconds).

awsutil/rotate.go
```diff
@@ -62,11 +62,11 @@ func (c *CredentialsConfig) RotateKeys(opt ...Option) error {
 // CreateAccessKey creates a new access/secret key pair.
 //
 // Supported options: WithEnvironmentCredentials, WithSharedCredentials,
-// WithAwsSession, WithUsername
+// WithAwsSession, WithUsername, WithTimeout
```
jefferai, line 65: This is a bit nonintuitive because it suggests that the timeout is related to creating the key itself, e.g. a timeout on that actual creation call. It's probably worth adding an extra bit of commenting around what WithTimeout does here (so anyone looking at it on pkg.dev.go will see it, whereas they likely won't see the comment below).

awsutil/rotate.go
vancluever: Every second for a few seconds does not sound like a big deal! I'll make the change.

awsutil/rotate.go
vancluever: Done in latest commit!

vancluever:
@jefferai changes made, should be good to go!

PR #11, opened 2021-09-21
Read `AWS_PROFILE` to enable Shared Credentials provider

vinay-gopalan:
This PR uses the Profile field that already existed in the AWS Credentials Config. It reads from the `AWS_PROFILE` environment variable and uses this value in the `GenerateCredentialChain` method to enable the Shared Credentials Provider. 

The addition of these changes allows the use of `source_profile` fields that point to a named profile with credentials for assuming roles. This aims to fix:
- https://github.com/hashicorp/vault/issues/5767
- https://github.com/hashicorp/terraform-provider-vault/issues/1086 (TFVP issue related to above Vault issue)

### DIFF
--- a/awsutil/generate_credentials.go
+++ b/awsutil/generate_credentials.go
@@ -180,6 +180,14 @@ func (c *CredentialsConfig) GenerateCredentialChain(opt ...Option) (*credentials
 		providers = append(providers, webIdentityProvider)
 	}
 
+	profile := os.Getenv("AWS_PROFILE")
+	if profile != "" {
+		c.Profile = profile
+	}
+	if c.Profile == "" {
+		c.Profile = "default"
+	}
+
 	if opts.withEnvironmentCredentials {
 		// Add the environment credential provider
 		providers = append(providers, &credentials.EnvProvider{})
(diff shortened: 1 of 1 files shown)


awsutil/generate_credentials.go
```diff
@@ -180,6 +180,14 @@ func (c *CredentialsConfig) GenerateCredentialChain(opt ...Option) (*credentials
 		providers = append(providers, webIdentityProvider)
 	}
 
+	profile := os.Getenv("AWS_PROFILE")
```
jefferai, line 183: Since this is only used in the shared creds provider, can you move this into the `opts.withSharedCredentials` block below?

vinay-gopalan: @jefferai Fixed!

jefferai (approved):

jefferai:
Tagged `awsutil/v0.1.5`

jeffery:
How can I tell if this fix is part of vault release? I am currently on vault 1.9.1 and the issue reported in https://github.com/hashicorp/vault/issues/5767 is not fixed for me.

jeffery:
Hey @vinay-gopalan do you have a test case which this issue was determined to be fixed?

PR #13, opened 2021-10-21
Add day suffix support to ParseDurationSecond

kalafut:
Add support for a "d" suffix that corresponds to 24 hours. Also add more test cases and slightly refactor the code.

This type of support has been requested and rejected by the Go team (https://github.com/golang/go/issues/11473). Their push back was around "how long is a day?", but for our users I think defining a day as 24 hours is fine, and convenient for longer duration TTLs, e.g. root keys that are valid for 90 days.

### DIFF
--- a/parseutil/parseutil_test.go
+++ b/parseutil/parseutil_test.go
@@ -166,6 +166,7 @@ func Test_ParseDurationSecond(t *testing.T) {
 		// Numeric inputs
 		{in: 9876, out: 9876 * time.Second},
 		{in: 5.5, out: 5 * time.Second},
+		{in: 5, out: 5 * time.Second},
 		{in: 0.9, out: 0 * time.Second},
 		{in: -5, out: -5 * time.Second},
 
@@ -174,6 +175,7 @@ func Test_ParseDurationSecond(t *testing.T) {
 		{in: "9876s", out: 9876 * time.Second},
 		{in: "50ms", out: 50 * time.Millisecond},
 		{in: "0.5m", out: 30 * time.Second},
+		{in: "0.5s", out: 500 * time.Millisecond},
 		{in: "5m", out: 5 * time.Minute},
 		{in: "6h", out: 6 * time.Hour},
 		{in: "5d", out: 5 * 24 * time.Hour},


parseutil/parseutil.go
```diff
@@ -106,19 +106,28 @@ func ParseDurationSecond(in interface{}) (time.Duration, error) {
...
-		} else {
-			// Plain integer
-			secs, err := strconv.ParseInt(inp, 10, 64)
+		case "d":
+			// Parse days (which aren't supported by the stdlib) by stripping the suffix and adjusting the scale.
+			scale = 24 * time.Hour
+			inp = inp[:len(inp)-1]
+			fallthrough
+		default:
+			val, err := strconv.ParseFloat(inp, 64)
```
calvn, line 126: Want to call out that changing this to `strconv.ParseFloat` means that we now accept fractional values so it's a behavioral change.

kalafut: I only today learned that the suffix variants already supported fractional:

```
➜  ~ vault token create -ttl=5.5h
Key                  Value
---                  -----
token                s.rfMyxR65n7GZARo9A1wvdwFk
token_accessor       0kjRm43pa71ODDeY1KeY0KMd
token_duration       5h30m
token_renewable      true
token_policies       ["root"]
identity_policies    []
policies             ["root"]
```

So supporting fractional for days seemed logical, but you're right that is does change the parsing for the no-suffix number.

parseutil/parseutil.go
```diff
@@ -106,19 +106,28 @@ func ParseDurationSecond(in interface{}) (time.Duration, error) {
...
 			if err != nil {
 				return dur, err
 			}
-			dur = time.Duration(secs) * time.Second
+			dur = time.Duration(val * float64(scale))
```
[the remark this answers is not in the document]
kalafut: I'm not following? Under the hood they're all ns precsion. The conversion to float64 is needed otherwise you can have 0.5 * int64(3000000) = 0

parseutil/parseutil.go
kalafut: Actually, ParseFloat introduces a bunch of unwanted edgecases. I'll revert to ParseInt and we'll only allow integral days.

parseutil/parseutil.go
```diff
@@ -106,19 +106,28 @@ func ParseDurationSecond(in interface{}) (time.Duration, error) {
 			return dur, nil
 		}
 		var err error
-		// Look for a suffix otherwise its a plain second value
-		if strings.HasSuffix(inp, "s") || strings.HasSuffix(inp, "m") || strings.HasSuffix(inp, "h") || strings.HasSuffix(inp, "ms") {
+
+		suffix := inp[len(inp)-1:]
```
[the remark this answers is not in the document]
kalafut: I think this will be OK as is. All [UTF-8 encoded Unicode](https://en.wikipedia.org/wiki/UTF-8#Encoding) are made up of bytes >= 0x80, so if part of a multi-byte character is sliced off the remaining string will have a high byte that will fail to parse.

parseutil/parseutil.go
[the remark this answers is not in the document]
kalafut: Resolving this this has been removed.

parseutil/parseutil.go
```diff
@@ -106,19 +106,28 @@ func ParseDurationSecond(in interface{}) (time.Duration, error) {
 			return dur, nil
 		}
 		var err error
-		// Look for a suffix otherwise its a plain second value
-		if strings.HasSuffix(inp, "s") || strings.HasSuffix(inp, "m") || strings.HasSuffix(inp, "h") || strings.HasSuffix(inp, "ms") {
+
```
ncabatoff, line 109: Alternate approach to consider: try ParseInt first.  If that returns success, we don't have to worry about suffix.  If ParseInt fails, look for and handle a `d` suffix, failing which, use ParseDuration.

kalafut: I think you're approach is cleaner. 👍

PR #17, opened 2021-11-12
feature: Add support handling forwarded for headers 

jimlambrt:
- `TrustedFromXForwardedFor` uses the XForwardedFor* listener cfg setting to determine how/if headers are trusted/allowed for inbound http requests
- `WrapForwardedForHandler` which is an http middleware wrapper that uses `TrustedFromXForwardedFor` to determine how/if headers are trusted/allowed for inbound http requests

### DIFF
--- a/listenerutil/forwarded_for.go
+++ b/listenerutil/forwarded_for.go
@@ -15,6 +15,8 @@ type key int
 
 const (
 	remoteAddrKey key = iota
+
+	missingPortErrStr = "missing port in address"
 )
 
 // ErrResponseFn provides a func to call whenever WrapForwardedForHandler
@@ -93,6 +95,8 @@ func TrustedFromXForwardedFor(r *http.Request, l *ListenerConfig) (*Addr, error)
 		return nil, fmt.Errorf("missing x-forwarded-for header and configured to reject when not present")
 	}
 
+	// http request remote address will always have a host:port
+	// (see: https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/net/http/request.go;l=279-286)
 	host, port, err := net.SplitHostPort(r.RemoteAddr)
 	if err != nil {
 		// If not rejecting treat it like we just don't have a valid
@@ -133,20 +137,30 @@ func TrustedFromXForwardedFor(r *http.Request, l *ListenerConfig) (*Addr, error)
 
 	// Split comma separated ones, which are common. This brings it in line
 	// to the multiple-header case.
-	var acc []string
+	var acc []Addr
 	for _, header := range headers {
 		vals := strings.Split(header, ",")
 		for _, v := range vals {
 			// validate the header contains a valid IP
 			v = strings.TrimSpace(v)
-			ip := net.ParseIP(v)
+			h, p, err := net.SplitHostPort(v)
+			switch {
+			case err != nil && strings.Contains(err.Error(), missingPortErrStr):
+				h = v
+			case err != nil && !strings.Contains(err.Error(), missingPortErrStr):
+				if !rejectNotPresent {
+					return nil, nil
+				}
+				return nil, fmt.Errorf("error parsing client address host/port (%s) from header", v)
+			}
+			ip := net.ParseIP(h)
 			if ip == nil {
 				if !rejectNotPresent {
 					return nil, nil
 				}
 				return nil, fmt.Errorf("error parsing client address (%s) from header", v)
 			}
-			acc = append(acc, v)
+			acc = append(acc, Addr{Host: h, Port: p})
 		}
 	}
 
@@ -165,7 +179,11 @@ func TrustedFromXForwardedFor(r *http.Request, l *ListenerConfig) (*Addr, error)
 		return nil, fmt.Errorf("malformed x-forwarded-for configuration or request, hops to skip (%d) would skip before earliest chain link (chain length %d)", hopSkips, len(headers))
 	}
 
-	return &Addr{acc[indexToUse], port}, nil
+	// TO-DO: using the remote address port here is not correct and still needs to be
+	// resolved.  We could return the trusted port... but it could also be
+	// empty.  Looking for some guidance from reviewers, especially on the
+	// downstream vault affects if the port returned is empty
+	return &Addr{acc[indexToUse].Host, port}, nil
 }
 
 // newOrigRemoteAddrCtx will return a context containing a value for the
--- a/listenerutil/forwarded_for.go
+++ b/listenerutil/forwarded_for.go
@@ -40,12 +40,12 @@ func WrapForwardedForHandler(h http.Handler, l *ListenerConfig, respErrFn ErrRes
 	}
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 
-		trusted, err := TrustedFromXForwardedFor(r, l)
+		trusted, remoteAddr, err := TrustedFromXForwardedFor(r, l)
 		if err != nil {
 			respErrFn(w, http.StatusBadRequest, err)
 			return
 		}
-		if trusted == nil {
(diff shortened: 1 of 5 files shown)


jimlambrt:
@briankassouf @jefferai - I'm currently refactoring this PR quite a bit.  I'll ping you when it's ready for review.  ty!

jimlambrt:
@briankassouf @jefferai - okay, this is ready for review now.

listenerutil/forwarded_for.go
```diff
@@ -0,0 +1,192 @@
...
+
+	headers, headersOK := r.Header[textproto.CanonicalMIMEHeaderKey("X-Forwarded-For")]
+	if !headersOK || len(headers) == 0 {
+		if !rejectNotPresent {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("missing x-forwarded-for header and configured to reject when not present")
+	}
+
+	host, port, err := net.SplitHostPort(r.RemoteAddr)
```
jefferai, line 96: I'm not sure if RemoteAddr will always contain a port; if not, be aware that SplitHostPort errors if there is no port segment, so you may want to account for this.

listenerutil/forwarded_for.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// Split comma separated ones, which are common. This brings it in line
+	// to the multiple-header case.
+	var acc []string
+	for _, header := range headers {
+		vals := strings.Split(header, ",")
+		for _, v := range vals {
+			// validate the header contains a valid IP
+			v = strings.TrimSpace(v)
+			ip := net.ParseIP(v)
```
jefferai, line 142: In contrast to above, this I think will error if you _do_ have a port

jefferai (commented):
I think it looks good with the caveats around the host/port parsing in my comments. Also, I thought you wanted to handle X-Real-Ip or the other one too? Or is it not worth doing it all in the same place?

listenerutil/forwarded_for.go
jimlambrt: It doesn't error if there's no port: https://go.dev/play/p/jM1UsbsMXtH

listenerutil/forwarded_for.go
jimlambrt: I agree that it won't always contain a port.  In this case, an error will be raised and handled based on the value of `rejectNotPresent`.  
This mimics how vault handles this case: https://github.com/hashicorp/vault/blob/87c355e2bb34018b03e886f2e9155a6b3aba69fb/http/handler.go#L488
Do we want to change how it's handled?

jimlambrt:
I've intentionally left out the realIP bits, so the handling mimics the current vault implementation.  In Boundary, we'll use the return from `TrustedFromXForwardedFor` and then layer on how we want to handle realIP.

listenerutil/forwarded_for.go
jefferai: Yes, I think so. Looking at MDN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For in all of their examples no port is present. Even the examples for the standardized version of it (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) don't have ports shown.

Probably Vault has this wrong. @briankassouf curious your thoughts around this.

listenerutil/forwarded_for.go
jefferai: Right, but it does if you do have one! So the only way to get multiple addresses to be checked successfully would be for the first value to have a port and all subsequent values to not have ports. I doubt there's any proxy or LB that behaves that way.

I think in all cases we probably need to pass through `net.SplitHostPort` and either use the host value if a port was split or if we got an error about a missing port (I think it is sadly still not exported so it'd have to be a string check) we'd use the bare address.

Would love to get input from @briankassouf or @ncabatoff here.

jefferai:
> I've intentionally left out the realIP bits, so the handling mimics the current vault implementation. In Boundary, we'll use the return from TrustedFromXForwardedFor and then layer on how we want to handle realIP.

Might be worth implementing them here as optional bits that can be activated. Same if we want to handle `Forwarded` although I'm not sure how common that actually is.

listenerutil/forwarded_for.go
briankassouf: This is parsing the RemoteAddr that is provided in the http.Request object, which, may always have a port? https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/net/http/request.go;l=279-286

briankassouf: I agree X-Forwarded-For headers may not have a port, but i don't think we have that invariant below when parsing those values

listenerutil/forwarded_for.go
jimlambrt: Ah.. I didn't realize that x-forwarded-for could have ports as well. ty!

listenerutil/forwarded_for.go
jefferai: Oh, I'm sorry. In this re-review I mixed up RemoteAddr vs X-Forwarded-For as you can clearly see in my above comments :-)

listenerutil/forwarded_for.go
jefferai: I don't know if they can or not. Maybe they never can. I was mixing up RemoteAddr above vs. X-Forwarded-For.

jefferai (dismissed):

listenerutil/forwarded_for.go
jimlambrt: I believe that it may have a port: https://www.rfc-editor.org/rfc/rfc7239#section-5.2

jefferai: That's Forwarded which is different -- an attempt to standardize the defacto X-Forwarded-For.

It won't _hurt_ if you want to detect port and use it if it's there. Then the replaced RemoteAddr can have a port too.

briankassouf: We also stash the original port number and merge it back to the RemoteAddr, so there is another bit of code here assuming there will not be port number

jimlambrt: So... I think, we need to change how port is handled a bit.  This code and the current vault code, takes the trusted `acc` the remote address' port and assumes they belong together.  In vault this could result in the request remote address having a port that belongs to the inbound remote address but an IP that comes from a trusted x-forwarded-for header.

jimlambrt: Hum... the more I ponder this code, the more curious I get about port.  It would appear that a request's remote address must have a port... since x-forwarded-for headers aren't required to have a port, there seems to be a misalignment here when vault overwrites the remote address with the trusted header.

jimlambrt: @jefferai @briankassouf -> pushed a commit to address most of the comment I believe.. but still need to resolve how we want to handle port, especially when remote addr is overwritten with it.

briankassouf: Did we ever figure out if X-Forwarded-For may include port numbers? Unfortunately it's not clear to me which port number we ought to use if there are two different ones. It probably makes sense to always ensure a port is set though so we don't break the assumption down the call chain that `RemoteAddr` always contains a port

jimlambrt: Based on some googling, It would appear that it's not uncommon it to have a port if folks are using IIS servers running ARR.

jimlambrt: Okay... pushed another commit which is a possible solution based on this discussion and it's ready for re-review. ty!

jefferai (approved):
Looks good!

jimlambrt:
@briankassouf -> I'm going to merge but feel free to add additional comments to the PR and I'll address them in a new PR.

PR #19, opened 2021-11-17
Minimize allocations in strutil

sethvargo:
I noticed some pretty quick-and-easy wins for performance in the strutil package.

1. Use `struct{}` instead of `bool` for map membership. The empty struct does not allocate and consumes no memory, so it's far more efficient than a boolean:

    ```go
    package main
    
    import (
    	"fmt"
    	"unsafe"
    )
    
    func main() {
    	x := [100]struct{}{}
    	y := [100]bool{}
    
    	fmt.Printf("%v\n", unsafe.Sizeof(x)) // 0
    	fmt.Printf("%v\n", unsafe.Sizeof(y)) // 100
    }
    ```

1. Pre-allocate slices and maps when we know the maximum possible size.

1. Remove an unnecessary test dependency and rename the tests to match Go's expected test format name.

## Benchmarks

Before:

```text
BenchmarkRemoveDuplicates-16          	       2	 756790682 ns/op	106821536 B/op	   38192 allocs/op
BenchmarkRemoveDuplicatesStable-16    	       6	 183990038 ns/op	58343824 B/op	       4 allocs/op
BenchmarkEquivalentSlices-16          	       1	1666107197 ns/op	358676792 B/op	   76680 allocs/op
```

After:

```text
BenchmarkRemoveDuplicates-16          	       3	 477817660 ns/op	56117237 B/op	       7 allocs/op
BenchmarkRemoveDuplicatesStable-16    	       6	 200917214 ns/op	56115224 B/op	       3 allocs/op
BenchmarkEquivalentSlices-16          	       1	1009316412 ns/op	112230496 B/op	       8 allocs/op
```

### DIFF
--- a/strutil/strutil.go
+++ b/strutil/strutil.go
@@ -230,16 +230,16 @@ func TrimStrings(items []string) []string {
 // strings. This also may convert the items in the slice to lower case and
 // returns a sorted slice.
 func RemoveDuplicates(items []string, lowercase bool) []string {
-	itemsMap := map[string]bool{}
+	itemsMap := make(map[string]struct{}, len(items))
 	for _, item := range items {
 		item = strings.TrimSpace(item)
-		if lowercase {
-			item = strings.ToLower(item)
-		}
 		if item == "" {
 			continue
 		}
-		itemsMap[item] = true
+		if lowercase {
+			item = strings.ToLower(item)
+		}
+		itemsMap[item] = struct{}{}
 	}
 	items = make([]string, 0, len(itemsMap))
 	for item := range itemsMap {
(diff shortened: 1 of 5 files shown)


strutil/strutil.go
```diff
@@ -230,16 +230,16 @@ func TrimStrings(items []string) []string {
 // strings. This also may convert the items in the slice to lower case and
 // returns a sorted slice.
 func RemoveDuplicates(items []string, lowercase bool) []string {
-	itemsMap := map[string]bool{}
+	itemsMap := make(map[string]struct{}, len(items))
 	for _, item := range items {
 		item = strings.TrimSpace(item)
-		if lowercase {
```
sethvargo, line 236: Moved this lower, because it didn't seem like a good idea to try and lowercase the empty string.

sethvargo:
Pokey @jefferai @calvn

jefferai:
I'm on vacation :-)

jefferai (approved):

jefferai:
Thanks!

sethvargo:
Thank you 😄

PR #22, opened 2021-11-29
chore: Add CONTRIBUTING.md

jimlambrt:


jefferai (commented):
Contributing to dbw, eh?

Also I think it's worth adding something up top to make it clear that these are actively cross project and to ensure leads from all affected projects are aware of and given a chance to review.

I realize all libraries are cross project but this is fundamental stuff around configuring system variables and listeners and parsing...

jefferai (approved):

PR #25, opened 2021-12-10
feat(listener): New TLS fields to support env and file sourcing

hugoamvieira:
The deprecated fields, `TLSKeyFile` and `TLSClientCAFile` are limited in terms of functionality: They only accept a path to a file containing their respective values.

The newly introduced fields `TLSKey` and `TLSClientCA` are more versatile: They abstract away the file limitation and contain the actual values. You can set these values directly, source them from an environment variable using the `env://` URL, or from a file using the `file://` URL.


hugoamvieira:
d5f6233 Adds checks to ignore `ErrNotAUrl` since it is possible to set the values directly in the HCL. This logic will pass that into `ParsePath` and it might result in this error on an otherwise valid value.

jefferai (approved):

PR #29, opened 2022-03-03
Handle JSON Number in ParseCommaStringSlice

ccapurso:
The `ParseCommaStringSlice` function in `parseutil` performs a type assertion (i.e. `in.(string)`) but does not return an error if the assertion fails. Without an early return, a panic might occur in the `mapstructure.NewDecoder` call. `StringToSliceHookFunc` in `mapstructure` returns a hook func that performs a similar string type assertion but [does not verify that it was successful](https://github.com/mitchellh/mapstructure/blob/1b7c3d8e219d00abb1d9267545c50c6c12fee562/decode_hooks.go#L91). This can occur if the input is a `json.Number` given that its `reflect.Kind` is `string`. The type assertion in `mapstructure` will panic with `invalid type assertion: data.(string) (non-interface type json.Number on left)`.

Unit tests for `ParseCommaStringSlice` were also added as there weren't any prior to this PR.

### DIFF
--- a/parseutil/parseutil_test.go
+++ b/parseutil/parseutil_test.go
@@ -2,6 +2,7 @@ package parseutil
 
 import (
 	"encoding/json"
+	"math/cmplx"
 	"testing"
 	"time"
 )
@@ -461,12 +462,12 @@ func Test_ParseIntSlice(t *testing.T) {
 		{
 			"1,2",
 			true,
-			[]int64{1,2},
+			[]int64{1, 2},
 		},
 		{
 			"1,2,3",
 			true,
-			[]int64{1,2,3},
+			[]int64{1, 2, 3},
 		},
 	}
 
@@ -503,12 +504,18 @@ func equalStringSlice(a, b []string) bool {
 }
 
 func Test_ParseCommaStringSlice(t *testing.T) {
-	cases := []struct{
+	cases := []struct {
 		name     string
-		inp       interface{}
+		inp      interface{}
 		expected []string
 		valid    bool
 	}{
+		{
+			"nil",
+			nil,
+			[]string{},
+			true,
+		},
 		{
 			"empty string",
 			"",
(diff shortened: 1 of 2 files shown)


parseutil/parseutil_test.go
```diff
@@ -487,3 +487,85 @@ func Test_ParseIntSlice(t *testing.T) {
 		}
 	}
 }
+
+func equalStringSlice(a, b []string) bool {
```
ncabatoff, line 491: There's a helper for this in strutil if you want.

ncabatoff (dismissed):

parseutil/parseutil_test.go
ccapurso: The one similar helper that I have found is `EquivalentSlices`. It doesn't doesn't actually check strict equality. Instead, it ensures that the sets of values are the same independent of the order. It ultimately performs a sort under the hood prior to looping and comparing each value. Were you thinking of a different function that I might have missed?

ncabatoff: Nope, just mistaken.

ccapurso: Bummer, thanks for the suggestion though! I'll keep that package in mind for the future.

briankassouf (commented):
I believe that prior to this change non-strings were valid inputs: https://go.dev/play/p/oy2jtFp-k7z

Do we instead want to special case json.Number?

ccapurso:
> I believe that prior to this change non-strings were valid inputs: https://go.dev/play/p/oy2jtFp-k7z
> 
> Do we instead want to special case json.Number?

Oh wow, this function is _a lot_ more permissive than I expected. I imagine that is due to setting `WeaklyTypedInput` to `true` in the `mapstructure.DecoderConfig`. For example, passing `[]bool{true, false, true}` results in `[]string{"1", "0", "1"}`. It's kind of interesting that this function has somewhat undefined/magical behavior based on how `WeaklyTypedInput` is set to `true`. Based on this, it definitely makes sense to just special case `json.Number`. I can expand the unit tests but do we actually intend to make this function this permissive? The name definitely doesn't imply that.

briankassouf:
@ccapurso Since we use it in our API parameter parsing code, i think it's hard to say if making it less permissive would break existing APIs usage or not. I'm also not sure if boundary is using it in any meaningful way. Since it takes an `interface{}` as input it's probably best to not break the existing contract and just special case json.Number

ccapurso:
@briankassouf, definitely agree. I didn't mean to imply that we should be less permissive. I am trying to get a feel for whether there might be other cases that might panic so that we can handle accordingly and ensure that there are appropriate tests.

briankassouf (approved):

PR #30, opened 2022-03-04
Add pluginutil

jefferai:
This adds pluginutil, which is used as common functionality for instantiating different types of plugins (currently, in Boundary, both dynamic host catalogs and KMS plugins).

Given the nature of testing with plugins and filesystem abstractions, testing is complicated. It's also a bit of a "works or doesn't" situation -- running Boundary spins out KMS and host catalog plugins with this code, and if it wasn't working, it'd be obvious -- you either have a working plugin or you don't. As a result, currently there are only tests on the options. 

(In the gkw-v2-testing branch in Boundary both in-memory and go-plugin based plugins are used, so both are exercised there.)


mgaffney (approved):

pluginutil/pluginutil.go
```diff
@@ -0,0 +1,197 @@
...
+// *<go-plugin>.Client or the value returned from an in-memory function. A type
+// switch should be used by the calling code to determine this, and the
+// appropriate service should be Dispensed if what is returned is a go-plugin
+// plugin.
+//
+// If the WithSecureConfig option is passed, this will be round-tripped into the
+// PluginClientCreationFunction from the given *PluginInfo, where it can be sent
+// into the go-plugin client configuration.
+func CreatePlugin(plugin *PluginInfo, opt ...Option) (interface{}, func() error, error) {
+	switch {
```
jimlambrt, line 100: what happens if plugin is nil?

pluginutil/pluginutil.go
```diff
@@ -0,0 +1,197 @@
...
+		return nil, nil, fmt.Errorf("no inmem creation func and either filename or plugin creation func not provided")
+	}
+
+	opts, err := GetOpts(opt...)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error parsing plugin options: %w", err)
+	}
+
+	// Open and basic validation
+	file, err := plugin.ContainerFs.Open(plugin.Filename)
```
jimlambrt, line 115: should we check if plugin.ContainerFs is nil?

pluginutil/pluginutil.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	stat, err := file.Stat()
```
jimlambrt, line 119: perhaps defer file.Close() here (and remove it below)?

pluginutil/pluginutil.go
```diff
@@ -0,0 +1,197 @@
...
+		uncompBuf := new(bytes.Buffer)
+		_, err = uncompBuf.ReadFrom(gzipReader)
+		gzipReader.Close()
+		if err != nil {
+			return nil, nil, fmt.Errorf("error reading gzip compressed data from reader: %w", err)
+		}
+		buf = uncompBuf.Bytes()
+	}
+
+	cleanup := func() error {
```
jimlambrt, line 157: nice.

pluginutil/pluginutil.go
```diff
@@ -0,0 +1,197 @@
...
+			for k, creationFunc := range sourceInfo.pluginMap {
+				pluginMap[k] = &PluginInfo{InmemCreationFunc: creationFunc}
+			}
+		}
+	}
+
+	return pluginMap, nil
+}
+
+// CreatePlugin instantiates a given plugin either via an in-memory function or
```
jimlambrt, line 89: Would you mind documenting the need to call the cleanup() func returned... otherwise the caller has to read the code to figure out what's returned.

jimlambrt (changes requested):
Couple bits I'm curious about. otherwise looks great!

pluginutil/pluginutil.go
jefferai: Catastrophe, probably.

I'll fix that.

pluginutil/pluginutil.go
jefferai: We should!

pluginutil/pluginutil.go
jefferai: Sure

jimlambrt (approved):

PR #36, opened 2022-03-18
Allow plugins to be specified as direct file paths

jefferai:
This contains two items:

1. In pluginutil, allows a direct plugin path to be specified, along with a (required) checksum and (optional) hash method used to generate the checksum. This is via an option that then factors into building the plugin map, and some plugin launching changes to understand these types.
2. In configutil, support for this new feature for KMS plugins. Additionally, the ability to specify this information within KMS blocks in HCL.

Tagging @jimlambrt for review, @calvn and @sgmiller mostly for heads-up.

### DIFF
--- a/configutil/Makefile
+++ b/configutil/Makefile
@@ -1,0 +1,8 @@
+PLUGIN_TMP_DIR := $(shell mktemp -d)
+
+test-plugin:
+	go build -o "${PLUGIN_TMP_DIR}/aeadplugin" testplugins/aead/main.go
+	PLUGIN_PATH="${PLUGIN_TMP_DIR}/aeadplugin" go test -v -run TestFilePlugin
+
+
+.PHONY: test-plugin
--- a/configutil/file_plugin_test.go
+++ b/configutil/file_plugin_test.go
@@ -4,12 +4,12 @@ import (
 	context "context"
 	"crypto/sha256"
 	"encoding/hex"
+	"fmt"
 	"os"
 	"testing"
 
 	wrapping "github.com/hashicorp/go-kms-wrapping/v2"
 	"github.com/hashicorp/go-secure-stdlib/pluginutil/v2"
-	"github.com/kr/pretty"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/crypto/sha3"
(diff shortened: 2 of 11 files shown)


configutil/file_plugin_test.go
```diff
@@ -0,0 +1,168 @@
+package configutil
+
+import (
+	context "context"
```
johanbrandhorst, line 4: Why is this aliased?

configutil/kms.go
```diff
@@ -208,10 +268,35 @@ func configureWrapper(
...
 	}
 
+	// If the KMS block contained plugin file information, add it
+	pluginOpts := opts.withPluginOptions
+	switch {
+	case configKMS.PluginPath == "" && configKMS.PluginChecksum == "":
+	case configKMS.PluginPath == "" && configKMS.PluginChecksum != "":
+		return nil, nil, errors.New("plugin checksum specified in kms but plugin path empty")
+	case configKMS.PluginPath != "" && configKMS.PluginChecksum == "" && len(configKMS.pluginChecksumBytes) == 0:
+		return nil, nil, errors.New("plugin path specified in kms but plugin checksum empty")
```
johanbrandhorst, line 278: What do you think about
```suggestion
	case configKMS.PluginPath == "":
		if configKMS.PluginChecksum != "" {
			return nil, nil, errors.New("plugin checksum specified in kms but plugin path empty")
		}
	case configKMS.PluginChecksum == "" && len(configKMS.pluginChecksumBytes) == 0:
		return nil, nil, errors.New("plugin path specified in kms but plugin checksum empty")
```
instead?

configutil/testplugins/aead/main.go
```diff
@@ -0,0 +1,17 @@
...
+	gkwp "github.com/hashicorp/go-kms-wrapping/plugin/v2"
+	aead "github.com/hashicorp/go-kms-wrapping/v2/aead"
+)
+
+func main() {
+	if err := gkwp.ServePlugin(aead.NewWrapper()); err != nil {
+		fmt.Println("Error serving plugin", err)
+		os.Exit(1)
+	}
+	os.Exit(0)
```
johanbrandhorst, line 16: Is this necessary?

pluginutil/pluginutil.go
```diff
@@ -101,28 +156,50 @@ func BuildPluginMap(opt ...Option) (map[string]*PluginInfo, error) {
...
-		return nil, nil, fmt.Errorf("no inmem creation func and either filename or plugin creation func not provided")
-	}
+	// We need the client creation func to use once we've spun out the plugin
+	case plugin.PluginClientCreationFunc == nil:
+		return nil, nil, fmt.Errorf("plugin creation func not provided")
 
-	opts, err := GetOpts(opt...)
-	if err != nil {
-		return nil, nil, fmt.Errorf("error parsing plugin options: %w", err)
+	// Either we need to have either a validated FS to read from or secure config
```
johanbrandhorst, line 184:
```suggestion
	// Either we need to have a validated FS to read from or a secure config
```

configutil/testplugins/aead/main.go
jefferai: Maybe not. I copied this from elsewhere, where it was copied from elsewhere. I think whoever initially wrote it was trying to be explicit.

configutil/kms.go
jefferai: Either way. I kind of like when there are two vars to have that explicit possibility matrix enumerated in the cases because I feel like it makes it easier for others to at-a-glace see if all cases have been handled.

configutil/file_plugin_test.go
jefferai: Probably `go-imports` or a similar tool doing something weird.

jefferai, line 4:
```suggestion
	"context"
```

jefferai:
```suggestion
	"context"
```

configutil/kms.go
johanbrandhorst: That's definitely reasonable!

johanbrandhorst (approved):

pluginutil/pluginutil.go
```diff
@@ -101,28 +156,50 @@ func BuildPluginMap(opt ...Option) (map[string]*PluginInfo, error) {
...
+	// If we have a constructed filesystem, read from there
+	case plugin.ContainerFs != nil:
+		file, err = plugin.ContainerFs.Open(plugin.Path)
+		name = plugin.Path
+
+	// If we have secure config, read from disk
+	case plugin.SecureConfig != nil:
+		file, err = os.Open(plugin.Path)
+		name = filepath.Base(plugin.Path)
+
```
calvn, line 197: Would it be useful to check, perhaps as an option, whether the path provided is a symlink? In Vault's case we want to ensure that the plugin is actually coming from a specific directory. We perform such check in [here](https://github.com/hashicorp/vault/blob/e9adaffead120eb6ccdce76c2b401e7cd9013280/vault/plugin_catalog.go#L537-L543).

jefferai: Maybe?

In Vault there is a configured plugin directory, but additionally you must provide a shasum. It's unclear to me what the value is of disallowing symlinks so long as the sha matches, other than the fact that "plugin directory" is a thing.

This isn't a link to directories, it's to a specific file. If you want to have that be a symlink that can be pointed at different versions of plugins, e.g. for upgrades, I don't think that's an issue so long as the hash matches -- that's the actual security. Even more than that, the checksum is evaluated _after_ the plugin is copied to the new temp location (as bytes, not just a filesystem copy which could possibly copy the symlink). So when we verify the plugin checksum it will definitely not be a symlink.

Note that given the nature of this file-specific construct, if you want to make use of it, nothing prevents you from evaluating symlinks prior to passing the path into this option. So I lean away from building in that functionality centrally given that the real security is from the hash.

configutil/Makefile
```diff
@@ -0,0 +1,8 @@
+PLUGIN_TMP_DIR := $(shell mktemp -d)
+
+test-plugin:
```
jimlambrt, line 3: ty!

jimlambrt (approved):
ty!

PR #37, opened 2022-03-21
feat(parseutil) Add Safe variants of ParseInt*

cipherboy:
These proposed variants allow parsing smaller data types (such as `int`s)
from larger data types (the `int64` returned by `ParseInt{,Slice}(...)`),
validating that they are within the requested range prior to casting.
With the `SafeParseIntRange(...)` helper, we also allow validation of the
maximum expected number of elements in the slice.

`Signed-off-by: Alexander Scheel <alex.scheel@hashicorp.com>`

--

https://hashicorp.atlassian.net/browse/VAULT-3892 has some context around this change for those with access.

### DIFF
--- a/parseutil/parseutil.go
+++ b/parseutil/parseutil.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
 	"regexp"
 	"strconv"
 	"strings"
@@ -92,6 +93,9 @@ func ParseCapacityString(in interface{}) (uint64, error) {
 	return cap, nil
 }
 
+// Parse a duration from an arbitrary value (a string or numeric value) into
+// a time.Duration; when units are missing (such as when a numeric type is
+// provided), the duration is assumed to be in seconds.
 func ParseDurationSecond(in interface{}) (time.Duration, error) {
 	var dur time.Duration
 	jsonIn, ok := in.(json.Number)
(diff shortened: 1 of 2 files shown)


sgmiller (dismissed):

cipherboy:
Added tests for the feature.

sgmiller (dismissed):

parseutil/parseutil.go
```diff
@@ -401,3 +402,54 @@ func ParseAddrs(addrs interface{}) ([]*sockaddr.SockAddrMarshaler, error) {
 
 	return out, nil
 }
+
+func SafeParseIntRange(in interface{}, min int64, max int64) (int64, error) {
```
jefferai, line 406: This is not a problem of your making, but it can be really, really confusing at this point to figure out which function to use from among many similar ones (for instance, would I use the direct variant, safe variant, or neither?). Can you please add comments to these functions explaining what they do? I realize it is relatively obvious but it would be good to declare intentions so at a glance people can figure out if it's the right one.

jefferai (dismissed):
LGTM but please add method comments

parseutil/parseutil.go
cipherboy: Generally the direct should only be used if the value for sure is a sliced value. If you're unsure (especially in an API where either a direct value or a list of values could be provided), I'd think its probably best to use `ParseIntSlice`, hence why that was given the shorter name over the `*Direct*` variant (and is inclusive of the latter's functionality).

There's no corresponding safe version of the `*Direct*` variant as I'm not sure I personally see much use in it, as it was mostly a helper towards the more permissive `ParseInt` version that someone else might find useful (given its a shared library).

I've added doc texts to all methods, clarifying that when the result of a non-Safe variant is called, the caller would have to manually validate range constraints to cast the result to a more restrictive numeric type, but that the Safe variants handle this for the caller.

cipherboy:
Thanks for the feedback, updated. :-)

briankassouf (approved):
Thanks for adding all the doc comments

sgmiller (approved):

PR #38, opened 2022-04-14
adding fileutil

tvoran:
Adds a fileutil package, mainly for the purpose of sharing the caching file reader from https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/main/caching_file_reader.go

### DIFF
--- a/fileutil/caching_file_reader.go
+++ b/fileutil/caching_file_reader.go
@@ -27,7 +27,7 @@ type CachingFileReader struct {
 
 type cachedFile struct {
 	// buf is the buffer holding the in-memory copy of the file.
-	buf string
+	buf []byte
 
 	// expiry is the time when the cached copy is considered stale and must be re-read.
 	expiry time.Time
@@ -41,14 +41,16 @@ func NewCachingFileReader(path string, ttl time.Duration) *CachingFileReader {
 	}
 }
 
-func (r *CachingFileReader) ReadFile() (string, error) {
+func (r *CachingFileReader) ReadFile() ([]byte, error) {
 	// Fast path requiring read lock only: file is already in memory and not stale.
 	r.l.RLock()
 	now := r.currentTime()
 	cache := r.cache
 	r.l.RUnlock()
 	if now.Before(cache.expiry) {
-		return cache.buf, nil
+		newBuf := make([]byte, len(cache.buf))
+		copy(newBuf, cache.buf)
+		return newBuf, nil
 	}
 
 	// Slow path: read the file from disk.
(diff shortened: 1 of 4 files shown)


fileutil/caching_file_reader.go
```diff
@@ -0,0 +1,68 @@
...
+	r.l.Lock()
+	defer r.l.Unlock()
+
+	buf, err := ioutil.ReadFile(r.path)
+	if err != nil {
+		return "", err
+	}
+	r.cache = cachedFile{
+		buf:    string(buf),
+		expiry: now.Add(r.ttl),
```
tomhjp, line 64: It's very unlikely, and would require exceedingly high load, but we could theoretically spend an arbitrary amount of time waiting for the write lock at line 55, and so we could end up storing an expiry time that's already in the past, or at least more stale than it should be.

```suggestion
		expiry: r.currentTime().Add(r.ttl),
```

tomhjp (approved):

fileutil/caching_file_reader.go
tvoran: Agreed, thanks!

tomhjp (dismissed):

fileutil/caching_file_reader.go
```diff
@@ -0,0 +1,68 @@
...
+	r.l.RUnlock()
+	if now.Before(cache.expiry) {
+		return cache.buf, nil
+	}
+
+	// Slow path: read the file from disk.
+	r.l.Lock()
+	defer r.l.Unlock()
+
+	buf, err := ioutil.ReadFile(r.path)
```
swenson, line 58: nit: Since we're on Go 1.16, can we use `os.ReadFile`?

swenson (approved):
LGTM; small nits about removing `ioutil`

fileutil/caching_file_reader.go
tvoran: Yep, fixed in [5b61c29](https://github.com/hashicorp/go-secure-stdlib/pull/38/commits/5b61c292446dcaa6f7b50783fc501a72e50bd94a)

fileutil/caching_file_reader_test.go
```diff
@@ -0,0 +1,65 @@
+package fileutil
+
+import (
+	"io/ioutil"
```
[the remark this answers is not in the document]
tvoran: Fixed in [5b61c29](https://github.com/hashicorp/go-secure-stdlib/pull/38/commits/5b61c292446dcaa6f7b50783fc501a72e50bd94a)

calvn (approved):

tomhjp (approved):

PR #42, opened 2022-08-05
Add MustParsePath

jefferai:


parseutil/parsepath.go
```diff
@@ -40,7 +56,10 @@ func ParsePath(path string) (string, error) {
 		return strings.TrimSpace(string(contents)), nil
 	case "env":
 		return strings.TrimSpace(os.Getenv(strings.TrimPrefix(path, "env://"))), nil
+	default:
+		if mustParse {
+			return path, ErrNotParsed
```
louisruch, line 61: It seems the rest of this function behaves this way so ok to merge as is, but is there a reason we don't return `"", ErrNotParsed`

louisruch: NVM I see the comment

louisruch (dismissed):
LGTM

parseutil/parsepath.go
jefferai: Yeah because of the way it's intended to work with ErrNotAUrl I did that here too. I'm happy to change it, it seems more like  a "real" error condition.

PR #46, opened 2022-09-30
Cleanup file if execution dir provided

louisruch:


louisruch:
@jefferai PTAL found this while testing a TF provider issue

PR #49, opened 2022-11-02
Add Custom Headers To Config Parser

DanielleMiu:
adds two additional configuration settings to the `listener` config section:
`custom_api_response_headers` and `custom_ui_response_headers`

this update also adds `WrapCustomHeadersHandler` wrapper which includes all of the logic necessary to perform custom header injection automatically

### DIFF
--- a/configutil/config_test.go
+++ b/configutil/config_test.go
@@ -53,6 +53,8 @@ func TestParseConfig(t *testing.T) {
 				custom_api_response_headers {
 					"default" = {
 						"test" = ["default value", "default value 2"]
+						// try unsetting this one
+						"x-content-type-options" = []
 					}
 					"200" = {
 						"test" = ["200 value"]
@@ -70,7 +72,8 @@ func TestParseConfig(t *testing.T) {
 				custom_ui_response_headers {
 					"default" = {
 						"test" =          ["ui default value"]
-						"cache-control" = ["max-age=604800"]
+						// try overwriting this one
+						"CACHE-CONTROL" = ["max-age=604800"]
 					}
 					"200" = {
 						"test" = ["ui 200 value"]
@@ -94,7 +97,10 @@ func TestParseConfig(t *testing.T) {
 							"custom_api_response_headers": []map[string]interface{}{
 								{
 									"default": []map[string]interface{}{
-										{"test": []interface{}{"default value", "default value 2"}},
+										{
+											"test":                   []interface{}{"default value", "default value 2"},
+											"x-content-type-options": []interface{}{},
+										},
 									},
 									"200": []map[string]interface{}{
 										{"test": []interface{}{"200 value"}},
@@ -115,7 +121,7 @@ func TestParseConfig(t *testing.T) {
 									"default": []map[string]interface{}{
 										{
 											"test":          []interface{}{"ui default value"},
-											"cache-control": []interface{}{"max-age=604800"},
+											"CACHE-CONTROL": []interface{}{"max-age=604800"},
 										},
 									},
 									"200": []map[string]interface{}{
@@ -133,46 +139,45 @@ func TestParseConfig(t *testing.T) {
 								},
 							},
 						},
-						CustomApiResponseHeaders: map[int]map[string]string{
+						CustomApiResponseHeaders: map[int]map[string][]string{
 							0: {
-								"Test":                      "default value; default value 2",
-								"Content-Security-Policy":   "default-src 'none'",
-								"X-Content-Type-Options":    "nosniff",
-								"Strict-Transport-Security": "max-age=31536000; includeSubDomains",
-								"Cache-Control":             "no-store",
+								"Test":                      {"default value", "default value 2"},
+								"Content-Security-Policy":   {"default-src 'none'"},
+								"Strict-Transport-Security": {"max-age=31536000; includeSubDomains"},
+								"Cache-Control":             {"no-store"},
 							},
 							200: {
-								"Test": "200 value",
+								"Test": {"200 value"},
 							},
 							2: {
-								"Test": "2xx value",
+								"Test": {"2xx value"},
 							},
 							401: {
-								"Test": "401 value",
+								"Test": {"401 value"},
 							},
 							4: {
-								"Test": "4xx value",
+								"Test": {"4xx value"},
 							},
 						},
-						CustomUiResponseHeaders: map[int]map[string]string{
+						CustomUiResponseHeaders: map[int]map[string][]string{
 							0: {
-								"Test":                      "ui default value",
-								"Content-Security-Policy":   "default-src 'none'; script-src 'self'; frame-src 'self'; font-src 'self'; connect-src 'self'; img-src 'self' data:*; style-src 'self'; media-src 'self'; manifest-src 'self'; style-src-attr 'self'; frame-ancestors 'self'",
-								"X-Content-Type-Options":    "nosniff",
-								"Strict-Transport-Security": "max-age=31536000; includeSubDomains",
-								"Cache-Control":             "max-age=604800",
+								"Test":                      {"ui default value"},
+								"Content-Security-Policy":   {"default-src 'none'; script-src 'self'; frame-src 'self'; font-src 'self'; connect-src 'self'; img-src 'self' data:*; style-src 'self'; media-src 'self'; manifest-src 'self'; style-src-attr 'self'; frame-ancestors 'self'"},
+								"X-Content-Type-Options":    {"nosniff"},
+								"Strict-Transport-Security": {"max-age=31536000; includeSubDomains"},
+								"Cache-Control":             {"max-age=604800"},
 							},
 							200: {
-								"Test": "ui 200 value",
+								"Test": {"ui 200 value"},
 							},
 							2: {
-								"Test": "ui 2xx value",
+								"Test": {"ui 2xx value"},
 							},
 							401: {
-								"Test": "ui 401 value",
+								"Test": {"ui 401 value"},
 							},
 							4: {
-								"Test": "ui 4xx value",
+								"Test": {"ui 4xx value"},
 							},
 						},
 					},
--- a/configutil/config_test.go
+++ b/configutil/config_test.go
@@ -4,6 +4,8 @@ import (
 	"os"
 	"testing"
 
+	"github.com/hashicorp/go-secure-stdlib/listenerutil"
+
 	"github.com/stretchr/testify/require"
 )
 
@@ -44,6 +46,104 @@ func TestParseConfig(t *testing.T) {
 			expSharedConfig: nil,
 			expErr:          true,
 			expErrIs:        os.ErrNotExist,
+		},
+		{
+			name: "custom headers parsed and set correctly",
+			in: `
+			listener "tcp" {
+				custom_api_response_headers {
+					"default" = {
+						"test" = ["default value", "default value 2"]
+					}
+					"200" = {
+						"test" = ["200 value"]
+					}
+					"401" = {
+						"test" = ["401 value"]
+					}
+				}
+				custom_ui_response_headers {
+					"default" = {
+						"test" =          ["ui default value"]
+						"cache-control" = ["max-age=604800"]
+					}
+					"200" = {
+						"test" = ["ui 200 value"]
+					}
+					"401" = {
+						"test" = ["ui 401 value"]
+					}
+				}
+			}`,
+			expSharedConfig: &SharedConfig{
+				Listeners: []*listenerutil.ListenerConfig{
+					{
+						Type: "tcp",
+						RawConfig: map[string]interface{}{
+							"custom_api_response_headers": []map[string]interface{}{
+								{
+									"default": []map[string]interface{}{
+										{"test": []interface{}{"default value", "default value 2"}},
+									},
+									"200": []map[string]interface{}{
+										{"test": []interface{}{"200 value"}},
+									},
+									"401": []map[string]interface{}{
+										{"test": []interface{}{"401 value"}},
+									},
+								},
+							},
+							"custom_ui_response_headers": []map[string]interface{}{
+								{
+									"default": []map[string]interface{}{
+										{
+											"test":          []interface{}{"ui default value"},
+											"cache-control": []interface{}{"max-age=604800"},
+										},
+									},
+									"200": []map[string]interface{}{
+										{"test": []interface{}{"ui 200 value"}},
+									},
+									"401": []map[string]interface{}{
+										{"test": []interface{}{"ui 401 value"}},
+									},
+								},
+							},
+						},
+						CustomApiResponseHeaders: map[string]map[string]string{
+							"default": {
+								"Test":                      "default value; default value 2",
+								"Content-Security-Policy":   "default-src 'none'",
+								"X-Content-Type-Options":    "nosniff",
+								"Strict-Transport-Security": "max-age=31536000; includeSubDomains",
+								"Cache-Control":             "no-store",
+							},
+							"200": {
+								"Test": "200 value",
+							},
+							"401": {
+								"Test": "401 value",
+							},
+						},
+						CustomUiResponseHeaders: map[string]map[string]string{
+							"default": {
+								"Test":                      "ui default value",
+								"Content-Security-Policy":   "default-src 'none'; script-src 'self'; frame-src 'self'; font-src 'self'; connect-src 'self'; img-src 'self' data:*; style-src 'self'; media-src 'self'; manifest-src 'self'; style-src-attr 'self'; frame-ancestors 'self'",
+								"X-Content-Type-Options":    "nosniff",
+								"Strict-Transport-Security": "max-age=31536000; includeSubDomains",
+								"Cache-Control":             "max-age=604800",
+							},
+							"200": {
+								"Test": "ui 200 value",
+							},
+							"401": {
+								"Test": "ui 401 value",
+							},
+						},
+					},
+				},
+			},
+			expErr: false,
 		},
 	}
 
--- a/configutil/config_test.go
+++ b/configutil/config_test.go
@@ -5,7 +5,6 @@ import (
 	"testing"
 
 	"github.com/hashicorp/go-secure-stdlib/listenerutil"
-
 	"github.com/stretchr/testify/require"
 )
 
@@ -58,8 +57,14 @@ func TestParseConfig(t *testing.T) {
 					"200" = {
 						"test" = ["200 value"]
 					}
+					"2xx" = {
+						"test" = ["2xx value"]
+					}
 					"401" = {
 						"test" = ["401 value"]
+					}
+					"4xx" = {
+						"test" = ["4xx value"]
 					}
 				}
 				custom_ui_response_headers {
@@ -70,8 +75,14 @@ func TestParseConfig(t *testing.T) {
 					"200" = {
 						"test" = ["ui 200 value"]
 					}
+					"2xx" = {
+						"test" = ["ui 2xx value"]
+					}
 					"401" = {
 						"test" = ["ui 401 value"]
+					}
+					"4xx" = {
+						"test" = ["ui 4xx value"]
 					}
 				}
 			}`,
@@ -88,8 +99,14 @@ func TestParseConfig(t *testing.T) {
 									"200": []map[string]interface{}{
 										{"test": []interface{}{"200 value"}},
 									},
+									"2xx": []map[string]interface{}{
+										{"test": []interface{}{"2xx value"}},
+									},
 									"401": []map[string]interface{}{
 										{"test": []interface{}{"401 value"}},
+									},
+									"4xx": []map[string]interface{}{
+										{"test": []interface{}{"4xx value"}},
 									},
 								},
 							},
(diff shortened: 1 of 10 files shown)


listenerutil/parse.go
```diff
@@ -387,3 +413,122 @@ func ParseSingleIPTemplate(ipTmpl string) (string, error) {
...
+	"1xx",
+	"2xx",
+	"3xx",
+	"4xx",
+	"5xx",
+}
+
+const strictTransportSecurity = "max-age=31536000; includeSubDomains"
+
+// parseCustomResponseHeaders takes a raw config values for the
```
johanbrandhorst, line 428:
```suggestion
// parseCustomResponseHeaders takes a raw config value for the
```

listenerutil/parse.go
```diff
@@ -387,3 +413,122 @@ func ParseSingleIPTemplate(ipTmpl string) (string, error) {
...
+	h := make(map[string]map[string]string)
+	// if r is nil, we still should set the default custom headers
+	if responseHeaders == nil {
+		h["default"] = map[string]string{"Strict-Transport-Security": strictTransportSecurity}
+		return h, nil
+	}
+
+	customResponseHeader, ok := responseHeaders.([]map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("response headers were not configured correctly. please make sure they're in a slice of maps")
```
johanbrandhorst, line 443: The context in which a user will see this error is if they misconfigure their HCL. We should adapt the error message. Per the [HCL definitions](https://github.com/hashicorp/hcl/blob/main/hclsyntax/spec.md), I think we should use "list".
```suggestion
		return nil, fmt.Errorf("response headers were not configured correctly. Please make sure they're in a list of maps")
```

listenerutil/parse.go
```diff
@@ -387,3 +413,122 @@ func ParseSingleIPTemplate(ipTmpl string) (string, error) {
...
+		h["default"] = make(map[string]string)
+	}
+	if _, ok := h["default"]["Strict-Transport-Security"]; !ok {
+		h["default"]["Strict-Transport-Security"] = strictTransportSecurity
+	}
+
+	return h, nil
+}
+
+// isValidStatusCode checking for status codes outside the boundary
```
johanbrandhorst, line 481: This could be a tiny bit confusing 😂.
```suggestion
// isValidStatusCode checks for status codes outside the allowed range
```

johanbrandhorst (changes requested):
I realise you probably meant to mark this a draft, but lets make sure we add tests before we merge this.

configutil/config_test.go
```diff
@@ -45,6 +47,104 @@ func TestParseConfig(t *testing.T) {
...
+						"test" =          ["ui default value"]
+						"cache-control" = ["max-age=604800"]
+					}
+					"200" = {
+						"test" = ["ui 200 value"]
+					}
+					"401" = {
+						"test" = ["ui 401 value"]
+					}
+				}
```
johanbrandhorst, line 76: Could we add some `2xx`, `3xx` etc?

listenerutil/listener.go
```diff
@@ -146,6 +146,7 @@ PASSPHRASECORRECT:
 				badCiphers = append(badCiphers, cipherStr)
 			}
 		}
+		// TODO: change this to not use a vault-specific error
```
johanbrandhorst, line 149: Is this happening :P? Maybe in a separate PR?

listenerutil/parse.go
```diff
@@ -387,3 +413,139 @@ func ParseSingleIPTemplate(ipTmpl string) (string, error) {
...
+	"3xx",
+	"4xx",
+	"5xx",
+}
+
+const strictTransportSecurity = "max-age=31536000; includeSubDomains"
+const xContentTypeOptions = "nosniff"
+const cacheControl = "no-store"
+const apiContentSecurityPolicy = "default-src 'none'"
+const uiContentSecurityPolicy = "default-src 'none'; script-src 'self'; frame-src 'self'; font-src 'self'; connect-src 'self'; img-src 'self' data:*; style-src 'self'; media-src 'self'; manifest-src 'self'; style-src-attr 'self'; frame-ancestors 'self'"
```
johanbrandhorst, line 430: Is there a reference for why we chose this as the default CSP?

listenerutil/parse.go
```diff
@@ -387,3 +413,139 @@ func ParseSingleIPTemplate(ipTmpl string) (string, error) {
...
+	}
+	return hvMap, nil
+}
+
+func parseHeaderValues(header interface{}) (string, error) {
+	var sl []string
+	if _, ok := header.([]interface{}); !ok {
+		return "", fmt.Errorf("headers must be given in a list of strings")
+	}
+	headerValList := header.([]interface{})
```
johanbrandhorst, line 536:
```suggestion
	headerValList, ok := header.([]interface{})
	if !ok {
		return "", fmt.Errorf("headers must be given in a list of strings")
	}
```

johanbrandhorst (changes requested):

listenerutil/listener.go
DanielleMiu: this is mostly for future reference, though it is such a small change that I *could* add it in this pr?

johanbrandhorst: Go for it IMO

listenerutil/parse.go
DanielleMiu: these are the headers that were defined in the ticket

johanbrandhorst: I couldn't find a reference in the ticket so w/e I guess 😛

listenerutil/handler.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+
+	// Setting the default headers first
+	if val, ok := sch[0]; ok {
+		setter(val)
+	}
+
+	// Then setting the generic hundred-level headers
+	if val, ok := sch[statusCode/100]; ok {
```
johanbrandhorst, line 61: Could you add a comment here noting that this uses integer division, which always rounds down?
```suggestion
	// Note: this integer division always rounds down, so 499/100 = 4
	if val, ok := sch[statusCode/100]; ok {
```

listenerutil/handler.go
```diff
@@ -0,0 +1,101 @@
...
+		setter(val)
+	}
+
+	// Finally setting the status-specific headers
+	if val, ok := sch[statusCode]; ok {
+		setter(val)
+	}
```
johanbrandhorst, line 68: What is the behavior if multiple different levels specify conflicting header values? Should we use `w.Header().Add` instead in the setter? What happens for duplicates?

listenerutil/handler.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+type uiRequestFunc func(*http.Request) bool
+
+// WrapCustomHeadersHandler wraps the handler to pass a custom ResponseWriter struct to all
+// later wrappers and handlers to assign custom headers by status code. This wrapper must
+// be the outermost wrapper to function correctly.
```
johanbrandhorst, line 75: The terminology for this is a HTTP middleware, lets use that in the API docs. Also, this describes the internal implementations details, when as a user I just want to know that it does what I want with the config.
```suggestion
// WrapCustomHeadersHandler wraps the handler, adding custom headers as defined in the
// ListenerConfig. The isUiRequest callback is used to determine whether an incoming request
// should get the UI specific headers or the generic API headers. This middleware must
// be the outermost middleware to function correctly in the case of another middleware returning
// early.
```

johanbrandhorst: Actually that reminds me, what happens if this middleware wraps a handler that returns without calling w.WriteHeader? Should we call w.WriteHeader for good measure before returning from here?

listenerutil/parse.go
```diff
@@ -387,3 +413,153 @@ func ParseSingleIPTemplate(ipTmpl string) (string, error) {
...
+const uiContentSecurityPolicy = "default-src 'none'; script-src 'self'; frame-src 'self'; font-src 'self'; connect-src 'self'; img-src 'self' data:*; style-src 'self'; media-src 'self'; manifest-src 'self'; style-src-attr 'self'; frame-ancestors 'self'"
+
+// parseCustomResponseHeaders takes raw config values for the "custom_ui_response_headers"
+// and "custom_api_response_headers". It makes sure the config entry is passed in as a map
+// of status code to a map of header name and header values. It verifies the validity of the
+// status codes, and header values. It also adds the default headers values for "Cache-Control",
+// "Strict-Transport-Security", "X-Content-Type-Options", and "Content-Security-Policy".
+func parseCustomResponseHeaders(responseHeaders interface{}, uiHeaders bool) (map[int]map[string]string, error) {
+	h := make(map[int]map[string]string)
+	// if r is nil, we still should set the default custom headers
```
johanbrandhorst, line 439:
```suggestion
	// if responseHeaders is nil, we still should set the default custom headers
```

listenerutil/parse.go
```diff
@@ -387,3 +413,153 @@ func ParseSingleIPTemplate(ipTmpl string) (string, error) {
...
+			return "", fmt.Errorf("found a non-string header value: %v", vh)
+		}
+		headerVal := strings.TrimSpace(vh.(string))
+		if headerVal == "" {
+			continue
+		}
+		sl = append(sl, headerVal)
+
+	}
+	s := strings.Join(sl, "; ")
```
johanbrandhorst, line 562: Instead of joining the strings like this, can we not keep the separate values as a slice and just add all of them with `Header().Add(v)`? This feels a bit ugly to be doing here.

listenerutil/handler.go
DanielleMiu: the latest call to `Header().Set()` overrides previous headers added the same way

listenerutil/handler.go
DanielleMiu: I have another comment addressing this in the code [here](https://github.com/hashicorp/go-secure-stdlib/pull/49/files#diff-edb1c099ae693e6b00aa24b159e9cd9cb6aca1bab17f5d168bf0ae5b90757138R28), but the default behavior of `ResponseWriter` is such that if a call has not been made to `WriteHeader` when `Write` is called, then `WriteHeader(200)` is called automatically

listenerutil/handler.go
johanbrandhorst: Is that the behavior we want? I could imagine saying you want both the 4xx and 401 headers to be merged into one.

listenerutil/handler.go
johanbrandhorst: Yes, but that will happen outside of our custom response writer, will it not?

listenerutil/parse.go
johanbrandhorst: I don't think either of those are good enough reasons to keep this TBH. We should operate on these as well typed values until we know we want to merge them, and at that point we shouldn't be doing it manually with `;`, we should use the library tools (`w.Header().Add`), IMO.

listenerutil/handler.go
DanielleMiu: handled offline, this is the behavior we want, since we need overrides to work

listenerutil/handler.go
DanielleMiu: if you check out the link I posted, you can see the logic is replicated in the response writer and is even tested for in the handler tests for the 200 responses

johanbrandhorst: My concern is what happens if `Write` isn't called but the inner middleware simply returns? It'd get all the way out of the middleware chain and then call `WriteHeader` on a response writer that isn't ours, no?

configutil/go.mod
```diff
@@ -2,6 +2,9 @@ module github.com/hashicorp/go-secure-stdlib/configutil/v2
 
 go 1.17
 
+// TODO: remove
+replace github.com/hashicorp/go-secure-stdlib/listenerutil => ../listenerutil
+
```
johanbrandhorst, line 7: Lets not forget to remove this before merge 😁

listenerutil/handler.go
```diff
@@ -0,0 +1,109 @@
...
+func WrapCustomHeadersHandler(h http.Handler, config *ListenerConfig, isUiRequest uiRequestFunc) http.Handler {
+	// TODO: maybe we should perform some preparsing here on the headers? check for duplicates,
+	// headers that aren't allowed, etc.
```
johanbrandhorst, line 82: Do you want to remove this comment? I would expect this sort of parsing to happen in the parser, not here.

listenerutil/handler_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+			resp := w.Result()
+			assert.Equal(t, tt.expStatus, resp.StatusCode)
+
+			for header, expected := range tt.expHeaders {
+				assert.Empty(t, cmp.Diff(expected, resp.Header.Values(header)))
+			}
+		})
+	}
+}
```
johanbrandhorst, line 243: Great tests!

johanbrandhorst (dismissed):

listenerutil/parse.go
```diff
@@ -387,3 +413,142 @@ func ParseSingleIPTemplate(ipTmpl string) (string, error) {
 		return "", fmt.Errorf("multiple addresses found (%q), please configure one", out)
 	}
 }
+
+const strictTransportSecurity = "max-age=31536000; includeSubDomains"
```
jefferai, line 417: Can you name these with a bit more definition? e.g. `defaultStrictTransportSecurityHeader`, `defaultXContentTypeOptionsHeader", `defaultCacheControlHeader`, etc.

listenerutil/parse.go
```diff
@@ -387,3 +413,142 @@ func ParseSingleIPTemplate(ipTmpl string) (string, error) {
...
+// and "custom_api_response_headers". It makes sure the config entry is passed in as a map
+// of status code to a map of header name and header values. It verifies the validity of the
+// status codes, and header values. It also adds the default headers values for "Cache-Control",
+// "Strict-Transport-Security", "X-Content-Type-Options", and "Content-Security-Policy".
+func parseCustomResponseHeaders(responseHeaders interface{}, uiHeaders bool) (map[int]map[string][]string, error) {
+	h := make(map[int]map[string][]string)
+	// if responseHeaders is nil, we still should set the default custom headers
+	if responseHeaders == nil {
+		h[0] = map[string][]string{
+			"Strict-Transport-Security": {strictTransportSecurity},
```
jefferai, line 433: Please make these header name values const.

listenerutil/parse.go
```diff
@@ -93,6 +94,12 @@ type ListenerConfig struct {
 	CorsAllowedOrigins                       []string    `hcl:"cors_allowed_origins"`
 	CorsAllowedHeaders                       []string    `hcl:"-"`
 	CorsAllowedHeadersRaw                    []string    `hcl:"cors_allowed_headers"`
+
+	// Custom Http response headers
+	CustomApiResponseHeaders    map[int]map[string][]string `hcl:"-"`
```
jefferai, line 99: There is an overall issue in the file which is that header names are case-insensitive but the code here won't treat it that way. Go handles this properly if you use `http.Header`. I think reading them in from HCL as strings is probably fine but once you are mixing and matching values below (which I think actually should not be done) then probably actions should take place on an http.Header.

listenerutil/parse.go
```diff
@@ -387,3 +413,142 @@ func ParseSingleIPTemplate(ipTmpl string) (string, error) {
...
+			if err != nil {
+				return nil, err
+			}
+
+			h[status] = headerVal
+		}
+	}
+
+	// setting default headers
+	if h[0] == nil {
```
jefferai, line 476: Up above if no values are set then you are setting defaults. At this point here you know that values have been set, but you are adding headers in if they were not set by the user anyway. This would make it impossible for users to _unset_ values we are setting by default that they don't want. I think this entire block probably should be removed.

listenerutil/parse.go
```diff
@@ -387,3 +413,142 @@ func ParseSingleIPTemplate(ipTmpl string) (string, error) {
...
+		}
+		if uiHeaders {
+			h[0]["Content-Security-Policy"] = []string{uiContentSecurityPolicy}
+		} else {
+			h[0]["Content-Security-Policy"] = []string{apiContentSecurityPolicy}
+		}
+		return h, nil
+	}
+
+	customResponseHeader, ok := responseHeaders.([]map[string]interface{})
```
jefferai, line 445: Rather than doing these checks here for `[]map[string]interface{}` consider using `mapstructure` to do the decoding for you. Not that this is wrong, but it saves some boilerplate. Probably whether it's worth it depends on if we'd want to allow weak decoding that mapstructure supports so you can e.g. type either `200` or `"200"`. Pros and cons to that...

listenerutil/parse.go
DanielleMiu: so this is an interesting one. I thought the same, but it seems somewhere (internal to the hcl parser?) the header names *are* normalized. you can see this happening in the test. `cache-control` values overwrites the `Cache-Control` value, and the `test` values are normalized to `Test`

jefferai: Ah, I didn't look at the function below -- it's happening in `parseHeaders` -- the `textproto.CanonicalMIMEHeaderKey` call. This is the same call that `http.Header` uses internally (at least for now).

It might be nice for `parseHeaders` to return an `http.Header` which is then what is sent around for further use by the system. So `Custom(Ui)ResponseHeaders` becomes `map[int]http.Header`. Thoughts?

listenerutil/parse.go
DanielleMiu: so I've added what I think is a happy medium. the defaults are still set if a user doesn't specify that specific header, but if the user specifies an empty header, it'll clear the defaults. this will also be reflected in the config test

listenerutil/parse.go
DanielleMiu: would there be any benefits to the hcl by implementing `mapstructure`? I would personally prefer to avoid using an external package unless there are explicit benefits to using it

listenerutil/parse.go
DanielleMiu: I think this would be the best way to do it if we had access to the internal workings of ResponseWriter, but the interface we're given only accepts headers in a string/string format, so it would be a bit harder to use. with the map we can just iterate over the key/value pairs

listenerutil/parse.go
jefferai: That makes sense to me

listenerutil/parse.go
jefferai: As I said, depends on if we want to allow weak decoding, and also handles a lot of type checking/reflect stuff for us...makes it easy to declare the structure you want and have it populated for you. `mapstructure` is used all over HC code (it's our lib) so if it's helpful it's a totally fine dep to bring in.

listenerutil/parse.go
DanielleMiu: if we can still iterate over it then that would work fine. I just meant that the ResponseWriter interface just gives us access to a single http.Header via ResponseWriter.Header() so we need to use Header.Add or Header.Set no matter what.

listenerutil/parse.go
DanielleMiu: so I've been reading through the [mapstructure docs](https://pkg.go.dev/github.com/mitchellh/mapstructure), and it seems really helpful for decoding known structures, but I'm not sure how it would help here, since we need to do additional parsing on all the values anyway via `convertStatusCode` and `parseHeaders`

johanbrandhorst (approved):
LGTM

listenerutil/parse.go
```diff
@@ -387,3 +414,157 @@ func ParseSingleIPTemplate(ipTmpl string) (string, error) {
...
+			headerVal, err := parseHeaders(headerValMap)
+			if err != nil {
+				return nil, err
+			}
+
+			h[status] = headerVal
+		}
+	}
+
+	// setting default headers
```
sgmiller, line 483: This could be a bit more generic, a for loop over pairs of header names/defaults for example, but not that important.

sgmiller (commented):
My only meta comment is that it seems difficult to prevent the default headers from being added if that was actually desired.

listenerutil/parse.go
DanielleMiu: this is done in steps so that more specific status codes can overwrite less specific ones. ex, you can set a header on the default, but the same header on 200, and the 200 header will be returned instead of the default

and regarding preventing defaults, this is also possible by setting empty headers in your config file: https://github.com/hashicorp/go-secure-stdlib/pull/49/files#diff-462be8440ab1a6db8af3debb42b464e02433a564223b166d96dfbfb5702fdfb7R57-R58

listenerutil/handler.go
jefferai: I also have this concern. Seems like you might want to say "all 2xx level status codes need header X but 201 also needs header Y".

You _could_ duplicate the headers in the config file between 2xx and 201, but I'm not sure why the current behavior is necessary for overrides to work.

listenerutil/listener.go
```diff
@@ -85,7 +85,7 @@ func TLSConfig(
 	if err := cg.Reload(); err != nil {
 		// We try the key without a passphrase first and if we get an incorrect
 		// passphrase response, try again after prompting for a passphrase
-		if errors.As(err, &x509.IncorrectPasswordError) {
+		if errors.Is(err, x509.IncorrectPasswordError) {
```
jefferai, line 88: What's the rationale for this change?

listenerutil/parse.go
jefferai: Parsing is separate from validation. Using mapstructure would allow people to specify status codes as either strings or ints. Right now you're forcing headers to be specified as strings, which may not be intuitive since HCL supports ints. I'm fine with it in this case because we also need to support 2xx style values, but just an FYI.

PR #51, opened 2022-11-29
Fix Default CSP img-src

DanielleMiu:
Replaces all instances of `data:*` with `data:` to correctly set the `img-src` policy of the default `content-security-policy`.

`data:*` does not correctly allow `data:image/png;base64,...` type inline images, but `data:` does


tmessi (approved):

jefferai (approved):

jefferai:
Tagged as 2.0.6!

PR #53, opened 2023-02-10
listenerutil: Support http.Flusher, http.Hijacker, http.Pusher

johanbrandhorst:
Using the new http.ResponseController in Go 1.20, we can easily add support for http.Flusher, http.Hijacker, http.Pusher and the SetReadDeadline and SetWriteDeadline methods to the ResponseWriter wrapper. This fixes some services that rely on this interface being implemented for their functionality.

In Go 1.19, just implement Flusher, Pusher and Hijacker explicitly.

### DIFF
--- a/listenerutil/handler.go
+++ b/listenerutil/handler.go
@@ -4,15 +4,7 @@ import (
 	"net/http"
 )
 
-type ResponseWriter struct {
-	wrapped http.ResponseWriter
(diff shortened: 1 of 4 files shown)


jimlambrt (approved):
ty!

jimlambrt (approved):

PR #57, opened 2023-04-17
feat: support adding ec2-instance role provider

ddebko:
## Summary

This PR enables the ability to add ec2-instance role providers to the credential chain for awsutil.

### Changes

- Added the following options to support ec2-instance role provider: WithRoleExternalId, WithRoleTags, WithWebIdentityTokenFile, WithRoleSessionName, WithRoleArn
- GenerateCredentialChain() will append a ec2-instance role provider to the provider chain when roleARN is not empty and WebIdentityTokenFile is empty.

### DIFF
--- a/awsutil/generate_credentials.go
+++ b/awsutil/generate_credentials.go
@@ -41,7 +41,9 @@ type CredentialsConfig struct {
 
 	// If specified, the region will be provided to the config of the
 	// EC2RoleProvider's client. This may be useful if you want to e.g. reuse
-	// the client elsewhere.
+	// the client elsewhere. If not specified, the region will be determined
+	// by the environment variables "AWS_REGION" or "AWS_DEFAULT_REGION".
+	// Otherwise the default value is us-east-1.
 	Region string
 
 	// The filename for the shared credentials provider, if being used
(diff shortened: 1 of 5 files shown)


DanielleMiu (commented):
LGTM

awsutil/generate_credentials.go
```diff
@@ -100,6 +113,33 @@ func NewCredentialsConfig(opt ...Option) (*CredentialsConfig, error) {
 			}
 		}
 	}
+	c.RoleARN = opts.withRoleArn
+	if c.RoleARN == "" {
```
jefferai, line 117: One of the points of using this library is so that we can _not_ use env vars unless explicitly opted in, as they can provide a different identity and level of security than you might want an instance of a plugin to have. That's why only region uses the env var here, and the rest of these should not be used.

jefferai: Scratch that. Apparently these three have always allowed using the env var for some reason. 🤷

jefferai: I do think it would be best to toggle env var reading with the withEnvironmentCredentials option but that would be a potentially breaking change we'd have to synchronize with Vault so for now probably best to leave it alone.

jefferai (commented):
I can't test that this is correct or not but no issues that I know of/can see.

awsutil/generate_credentials.go
ddebko: What should happen when the user provides static credentials (WithAccessKey, WithSecretKey) and options for generating a Web Identity Role Provider? Should both provider types be added to the credential chain or should an error be return to the user?

jefferai: Both types...the credential chain ordering will prevail, but this allows passing backup if needed

jefferai: Fallthroughs are fickle. Better to just remove the switch.

ddebko: When testing the scenario of having both static credentials and a ec2-instance role provider credential, the static credentials took priority in the credential chain. The reason why I was able to determine this is because each credential provider had different policy access permissions. Is this the expected behavior of the credential chain?

fairclothjm (approved):
LGTM! I pulled in this change and ran the unit/acceptance tests for Vault AWS secrets and auth engines as well as a few manual tests.

PR #74, opened 2023-05-26
listenerutil: Add default response header options

irenarindos:


listenerutil/parse.go
```diff
@@ -436,21 +441,29 @@ const cacheControl = "Cache-Control"
 // of status code to a map of header name and header values. It verifies the validity of the
 // status codes, and header values. It also adds the default headers values for "Cache-Control",
 // "Strict-Transport-Security", "X-Content-Type-Options", and "Content-Security-Policy".
```
johanbrandhorst, line 443: Maybe add a comment here about supported options?

listenerutil/parse.go
```diff
@@ -110,8 +110,13 @@ func (l *ListenerConfig) GoString() string {
 	return fmt.Sprintf("*%#v", *l)
 }
 
-func ParseListeners(list *ast.ObjectList) ([]*ListenerConfig, error) {
+func ParseListeners(list *ast.ObjectList, opt ...Option) ([]*ListenerConfig, error) {
```
johanbrandhorst, line 113: Not your fault but can we add a comment to this function? Make sure to document the supported options.

listenerutil/options.go
```diff
@@ -0,0 +1,60 @@
...
+	return &opts, nil
+}
+
+// Option - how Options are passed as arguments
+type Option func(*options) error
+
+// options = how options are represented
+type options struct {
+	withDefaultResponseHeaders    map[int]http.Header
+	withDefaultApiResponseHeaders map[int]http.Header
```
johanbrandhorst, line 27: Do we need both of these too? I'd argue we should just add what we need for now.

listenerutil/parse.go
```diff
@@ -377,15 +382,15 @@ func ParseListeners(list *ast.ObjectList) ([]*ListenerConfig, error) {
...
+			customApiHeadersMap, err := parseCustomResponseHeaders(l.CustomApiResponseHeadersRaw, false, WithDefaultResponseHeaders(opts.withDefaultApiResponseHeaders))
 			if err != nil {
 				return nil, multierror.Prefix(fmt.Errorf("failed to parse custom_api_response_headers: %w", err), fmt.Sprintf("listeners.%d", i))
 			}
 			l.CustomApiResponseHeaders = customApiHeadersMap
 			l.CustomApiResponseHeadersRaw = nil
 
 			// if CustomUiResponseHeadersRaw is nil, we still need to set the default headers
-			customUiHeadersMap, err := parseCustomResponseHeaders(l.CustomUiResponseHeadersRaw, true)
+			customUiHeadersMap, err := parseCustomResponseHeaders(l.CustomUiResponseHeadersRaw, true, WithDefaultResponseHeaders(opts.withDefaultUiResponseHeaders))
```
johanbrandhorst, line 393: Using this option as implemented will overwrite both UI and API response headers - should we not ensure this option only overwrites the UI response headers?

johanbrandhorst (changes requested):

johanbrandhorst:
I'm commandeering this PR

johanbrandhorst (approved):
Looks mighty fine if I may say so myself.

johanbrandhorst:
I'm going to go ahead and merge this and we can come back to further comments later if needed.

listenerutil/options.go
```diff
@@ -0,0 +1,38 @@
...
+	withDefaultUiContentSecurityPolicyHeader string
+}
+
+func getDefaultOptions() options {
+	return options{}
+}
+
+// WithDefaultUiContentSecurityPolicyHeader provides a default value for the UI listener's
+// Content-Security-Policy header.
+func WithDefaultUiContentSecurityPolicyHeader(cspHeader string) Option {
```
jefferai, line 33: Isn't this used to _override_ the default CSP value? In which case this isn't a `WithDefault`, it's `With`

johanbrandhorst: It overrides the default only, I think saying `With..` implies that it can't be changed by the user.

configutil/kms.go
```diff
@@ -211,28 +219,40 @@ func ParseKMSes(d string, opt ...Option) ([]*KMS, error) {
...
 // populate the actual KMSes once the HCL decoding has been performed
-func filterKMSes(list *ast.ObjectList, opt ...Option) ([]*KMS, error) {
+func filterKMSes(list *ast.ObjectList, maxKmsBlocks int) ([]*KMS, error) {
 	seals := new([]*KMS)
 
 	// opt is used after the WithMaxKmsBlocks option so that what a user passes
 	// in can override the defaults here
 	if o := list.Filter("hsm"); len(o.Items) > 0 {
-		if err := parseKMS(seals, o, "hsm", append([]Option{WithMaxKmsBlocks(2)}, opt...)...); err != nil {
+		maxHsmKmsBlocks := maxKmsBlocks
```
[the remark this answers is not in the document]
johanbrandhorst: We'll still need 3 variables, but we could avoid 2 if statements, true. I'll open a separate PR.

listenerutil/options.go
johanbrandhorst: I could make this `WithOverrideDefault`?

johanbrandhorst: Seems a bit tautological though

PR #78, opened 2023-06-01
awsutil: adding WebIdentityToken to CredentialsConfig{}

tvoran:
Adds field `WebIdentityToken` to `CredentialsConfig{}`, along with a simple TokenFetcher to use the contents of a token in the WebIdentityProvider, instead of the path to a token file. Useful when working with a token requested directly for an EKS/k8s service account configured with IRSA, for example.

Used in https://github.com/hashicorp/vault-secrets-operator/pull/235

### DIFF
--- a/awsutil/generate_credentials.go
+++ b/awsutil/generate_credentials.go
@@ -403,10 +403,10 @@ func stsSigningResolver(service, region string, optFns ...func(*endpoints.Option
 // FetchTokenContents allows the use of the content of a token in the
 // WebIdentityProvider, instead of the path to a token. Useful with a
 // serviceaccount token requested directly from the EKS/K8s API, for example.
-type FetchTokenContents string
+type FetchTokenContents []byte
 
 var _ stscreds.TokenFetcher = (*FetchTokenContents)(nil)
 
 func (f FetchTokenContents) FetchToken(_ aws.Context) ([]byte, error) {
-	return []byte(f), nil
+	return f, nil
 }
--- a/awsutil/generate_credentials.go
+++ b/awsutil/generate_credentials.go
@@ -74,7 +74,7 @@ type CredentialsConfig struct {
 
 	// The web identity token (contents, not the file path) to use with the web
 	// identity token provider
-	WebIdentityToken []byte
+	WebIdentityToken string
 
 	// The http.Client to use, or nil for the client to use its default
 	HTTPClient *http.Client
@@ -239,7 +239,7 @@ func (c *CredentialsConfig) GenerateCredentialChain(opt ...Option) (*credentials
 				// Add the web identity role credential provider
 				providers = append(providers, webIdentityProvider)
 			}
-		} else if len(c.WebIdentityToken) > 0 {
+		} else if c.WebIdentityToken != "" {
 			c.log(hclog.Debug, "adding web identity provider with token", "roleARN", c.RoleARN)
 			sess, err := session.NewSession()
 			if err != nil {
--- a/awsutil/generate_credentials.go
+++ b/awsutil/generate_credentials.go
@@ -74,7 +74,7 @@ type CredentialsConfig struct {
 
 	// The web identity token (contents, not the file path) to use with the web
 	// identity token provider
-	WebIdentityToken string
+	WebIdentityToken []byte
 
 	// The http.Client to use, or nil for the client to use its default
 	HTTPClient *http.Client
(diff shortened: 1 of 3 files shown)


awsutil/generate_credentials.go
```diff
@@ -230,6 +234,22 @@ func (c *CredentialsConfig) GenerateCredentialChain(opt ...Option) (*credentials
 				// Add the web identity role credential provider
 				providers = append(providers, webIdentityProvider)
 			}
+		} else if c.WebIdentityToken != "" {
```
tomhjp, line 237: The `WebIdentityTokenFile` option has functional options, a small bit of validation, and an environment variable associated with it. It feels like we should probably have the first two for `WebIdentityToken` too for consistency, and maybe the third? Although I'm slightly inclined to omit the environment variable given the expected consumption pattern as far as I understand it.

tomhjp (commented):
LGTM, just some small suggestions for consistency.

awsutil/generate_credentials.go
```diff
@@ -379,3 +399,14 @@ func stsSigningResolver(service, region string, optFns ...func(*endpoints.Option
...
+
+// FetchTokenContents allows the use of the content of a token in the
+// WebIdentityProvider, instead of the path to a token. Useful with a
+// serviceaccount token requested directly from the EKS/K8s API, for example.
+type FetchTokenContents string
+
+var _ stscreds.TokenFetcher = (*FetchTokenContents)(nil)
+
+func (f FetchTokenContents) FetchToken(_ aws.Context) ([]byte, error) {
+	return []byte(f), nil
```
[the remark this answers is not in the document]
tvoran: Gotcha, and should `CredentialsConfig.WebIdentityToken` be a `[]byte` as well? 

For the operator case I believe we will have to cast from string to []byte somewhere along the line, since the k8s API returns the token as a string, but that's easily done on the operator side.

awsutil/generate_credentials.go
```diff
@@ -230,6 +234,22 @@ func (c *CredentialsConfig) GenerateCredentialChain(opt ...Option) (*credentials
...
+			c.log(hclog.Debug, "adding web identity provider with token", "roleARN", c.RoleARN)
+			sess, err := session.NewSession()
+			if err != nil {
+				return nil, errors.Wrap(err, "error creating a new session to create a WebIdentityRoleProvider with token")
+			}
+			webIdentityProvider := stscreds.NewWebIdentityRoleProviderWithToken(sts.New(sess), c.RoleARN, c.RoleSessionName, FetchTokenContents(c.WebIdentityToken))
+
+			// Check if the webIdentityProvider can successfully retrieve
+			// credentials (via sts:AssumeRole), and warn if there's a problem.
+			if _, err := webIdentityProvider.Retrieve(); err != nil {
```
benashz, line 247: I see that `RetrieveCreds()` calls `GenerateCredentialChain()` every time, so does this mean we are doubling the number of credentials generated?

awsutil/generate_credentials.go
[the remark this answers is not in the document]
tomhjp: I was thinking `WebIdentityToken` and its functional option can still be a `string`, and Go will accept passing it as `FetchTokenContents(c.WebIdentityToken)` without any further casting. No strong preference if you prefer consistently treating it as `[]byte` all the way through though.

awsutil/generate_credentials.go
tomhjp: I like that we do this for consistency with the `WebIdentityTokenFile` option, but perhaps it would be nice to be able to optionally turn off this validation check. Maybe in a follow-on PR though to keep this one tightly scoped?

tomhjp (approved):
LGTM - I share Ben's thoughts on the dangling credentials, but given the pre-existing behaviour on the related option I'm hesitant to block this PR on resolving that question.

awsutil/generate_credentials.go
benashz: Given the fact that we are using this library in VSO, I feel like we should definitely provide an opt out from this sort of validation. The validation seems more geared towards the vault CLI use-case.

awsutil/generate_credentials.go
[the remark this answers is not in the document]
tvoran: Oh gotcha, I like that better, updated in 3bde4d9

awsutil/generate_credentials.go
tvoran: Yep, thanks for pointing that out. Updated with the option and validation in 02a10db9112506b25905a3a4eda8ce0d776eb9b4 and 3bde4d9

awsutil/generate_credentials.go
tvoran: Ok, I tried adding a WithSkipWebIdentityValidity option in 452aa5a6f488f459043678674fb24774e25ff9cf to skip the extra Retrieve() with the webIdentityProvider's. The AssumeRoleProvider looks like its doing some caching under the hood, so that's why the option is scoped to just the webIdentityProvider's.

tomhjp (approved):
LGTM

benashz (approved):
LGTM!

awsutil/generate_credentials.go
tvoran: I suspect the main reason the extra Retrieve() is called here is to help with debugging, since otherwise the aws sdk [doesn't give back error messages from failed providers](https://github.com/aws/aws-sdk-go/blob/v1.34.0/aws/credentials/chain_provider.go#L78) as it goes down the credentials chain if a subsequent provider succeeds. (Just something to keep in mind.)

Issue #81, opened 2023-07-24
User Agent Sanitization


NodyHub:
**Is your feature request related to a problem? Please describe.**
The consul licensing service (CLS) is considering to pipe the User Agent header from the request directly into the storage system.  The CLS itself does not further process the value. From a defence-in-depth perspective, a sanitisation of the value before storing is recommended. 

**Describe the solution you'd like**
Sanitization of the user agent by remove "unexpected" characters or replace them by a specified one. Existing PoC implementation is [here](https://github.com/NodyHub/go-useragent-sanitizer).

**Describe alternatives you've considered**
Let the code life in an own library. 

**Additional context**
As after a discussion about sanitization, the first proposed implementation was a deny-list approach, that replaces unexpected characters. Deny-lists tend to be not complete and might miss out disallowed characters. To reduce the chance of re-inventing the wheel across our products, having a security maintained standardization would be great.

NodyHub:
Closed after discussion in PR. Reasoning can be found here: https://github.com/hashicorp/go-secure-stdlib/pull/82#issuecomment-1663906217

PR #82, opened 2023-07-24
Extension of strutil library by `ReplaceEachNonMatchingRune`

NodyHub:
This PR adds a new function to the strutil library
- `ReplaceEachNonMatchingRune`: Which replaces each runes in a given string that does not match the given regex with a defined string

This PR is the preparation to address #81.


jefferai:
I believe you'll need to have an updated version of strutil before uasanitizer can be finalized -- any chance you can split this PR such that the strutil change is in here and uasanitizer in a new one? That way the strutil one can be merged and strutil can be tagged, then the uasanitizer PR can reflect that new tag in its dep.

strutil/strutil.go
```diff
@@ -511,3 +512,37 @@ func Reverse(in string) string {
 	}
 	return string(out)
 }
+
+// ReplaceNonMatcher replaces all characters within a given that do not match
```
jefferai, line 516:
```suggestion
// ReplaceNonMatcher replaces all characters within a given string that do not match
```

strutil/strutil.go
```diff
@@ -511,3 +512,37 @@ func Reverse(in string) string {
...
+// a given regular expression. The function returns the updated string and the
+// number of replacements
+func ReplaceNonMatcher(in, matcher string, replaceWith byte) (string, int) {
+	var adjustments int
+	work := []byte(in)
+	allowList := regexp.MustCompile(matcher)
```
jefferai, line 522: Please adjust the function signature to return an error and use Compile. Using MustCompile within this function is dangerous as it can trigger a panic; MustCompile is safest when run in an `init()` function so you panic at startup rather than at some point during runtime. Since the regex isn't known in advance it can't be moved to an `init()` function and this should instead return an error.

strutil/strutil.go
```diff
@@ -511,3 +512,37 @@ func Reverse(in string) string {
 	}
 	return string(out)
 }
+
+// ReplaceNonMatcher replaces all characters within a given that do not match
+// a given regular expression. The function returns the updated string and the
+// number of replacements
+func ReplaceNonMatcher(in, matcher string, replaceWith byte) (string, int) {
```
jefferai, line 519: It seems a bit odd to pass in a byte as the replacement character, since a string is represented by bytes but is actually a sequence of runes. So if you want a single character, you should probably pass in a rune.

Going a step further though, why not just allow passing in a string? A string of length one will act the same as this function, but you can replace with longer strings if you want.

Of course, if you do that, now you've just created https://pkg.go.dev/regexp#Regexp.ReplaceAllString and I do have to wonder if there is value in this function over just using that.

strutil/strutil.go
```diff
@@ -511,3 +512,37 @@ func Reverse(in string) string {
...
+	var adjustments int
+	work := []byte(in)
+	allowList := regexp.MustCompile(matcher)
+	for idx := range work {
```
jefferai, line 523: This function will break on many strings. The reason is that a string holds a sequence of bytes but since those bytes can contain unicode modifiers there is not a 1:1 mapping between a byte and an actual character in the string. As a result the right way to do this would be to keep the input string and range over it, which will give the starting indexes of the actual code points, as well as the rune itself. Then you should decide if the rune is what matches or doesn't match and rather than do an inline replacement use a string builder to build up the new string.

You can see the problems this creates at https://play.golang.com/p/3R-3kk-BO9I -- in the first example it iterates over the string and properly handles runes with multiple bytes. In the second example it properly iterates but by indexing you are seeing only the first unicode code point which is why the output is incorrect. In the third example, which matches what is happening in this function, you can see how it's not actually acting on the underlying string characters but rather a potentially partial representation of the character.

This matters for several reasons. One is that this is a generic string function package, and so while this may work for your use-case for user agents, it fails for general strings. Another is that this can actually return incorrect values because depending on the bytes that may make up a compound code point you may replace one byte of such a code point, leading the output string to have a changed value but not to the value you intended, at least not when treated as an actual UTF-8 string.

jefferai (changes requested):
This needs a lot of rework. I also think that the strutil functions may be completely unnecessary in favor of the existing replace functions in the stdlib's regexp package.

strutil/strutil.go
```diff
@@ -7,6 +7,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"regexp"
```
jefferai, line 10: There are no tests for the added functions in this file

strutil/strutil.go
NodyHub: Great call, will do so!

strutil/strutil.go
NodyHub: Make sense to me to go for a rune/string as type for `replaceWith`. I will adjust that. 

re https://pkg.go.dev/regexp#Regexp.ReplaceAllString : This function is intended to be the reverse way. So like replacing everything that does not match the regex. Leveraging `Regexp.ReplaceAllString` will transform the code to be much smoother. Will work on that 👌

strutil/strutil.go
NodyHub: @jefferai thanks a lot for this valuable feedback -- it's a blast! 

It's super interesting what you all got and I realize that I missed a lot of stuff.

NodyHub:
Let me go ahead, integrate your feedback and let's revisit the proposal again. 

I will exclude the `uasanitizer` package from this pr and update everything accordingly.

strutil/strutil.go
NodyHub: I added test cases for both functions in the same style as the other test were written. I took your example from https://play.golang.com/p/3R-3kk-BO9I as a base to generate the tests.

NodyHub:
Do to the fact that I made quiet a bunch of mistakes during proposal of that function, I am wondering if that might be a good indicator to have a reviewed version of that!? 

What do you think @jefferai?

strutil/strutil.go
```diff
@@ -511,3 +512,35 @@ func Reverse(in string) string {
 	}
 	return string(out)
 }
+
+// ReplaceNonMatcher replaces all characters within a given string that do not match
+// a given regular expression. The function returns the updated string and the
+// number of replacements
+func ReplaceNonMatcher(in, matcher, replaceWith string) (string, error) {
```
jefferai, line 519: It's not clear from the name/comment what this does; as the comment says it matches characters (really runes) in the given string and replaces with the given string, but someone could easily think that it's substring matching, or more specifically that it will replace all matches of the regex on the input string with the given value, without realizing that the matching will be rune-by-rune even though the replacement value is a string instead of a rune.

Possibly making this more confusing is that the user can control whether to match a single char or a string just based on the regex itself. So a user could pass in a regex matching a substring to this function, not realizing that it will never match anything because this does rune-by-rune.

My comment from before still stands -- I don't really think this function needs to exist. I think any usage of this function can be solved more flexibly by just using the stdlib `regexp.ReplaceAllString` and, if the caller needs, forming their regex to negate the character set they want. This function basically does the same thing except forces a rune-by-rune comparison, which AFAIK can be done with the regex anyways.

NodyHub: Well, thanks for the feedback @jefferai. I understand that you do not be 100% aligned with that description. If you have a better idea, feel free to adjust in a way so that it is more clear.

Background of this PR is to have a base that I/people can use in a sanitazation function that we can hand out as best practice. I got adviced by multiple members of team security to place the code in this repo. I am wondering why a security related function should not take place in this repo? What would be your recommendation? I do not want to create another repository like `go-secure-stdlib2`. 

Using `regexp.ReplaceAllString` is perfect to implement a deny-list and remove single character that seems to be bogus. Implementing a deny list is a security anti-pattern. A common best practice that is handed out from security is to implement allow lists. This motivation given, we need an inverted version of `regexp.ReplaceAllString` which replaces everything that does not match a given regex. Having such a base-function in the `strutils` would make implementing of allow-lists much easier.

I highly appreciate your feedback and it super valuable.

NodyHub:
Hey @jefferai what's your opinion on the last response/update? Another internal team want's to integrate a user agent sanitization and waits for the contribution for #81.

strutil/strutil.go
jefferai: I think my suggestion is exactly the opposite of what you are saying re: allow/deny list: by creating a regex that explicitly says what is allowed, and then saying "deny anything that isn't on this list", by doing it the way I suggested you are actually creating an allow list and then telling it to replace anything that doesn't match.

The current mechanism you're suggesting is to do the opposite: to create a regex that must match _everything_ that isn't allowed. If that problem space expands, you will no longer filter things that should be filtered.

Again, it's not that I think a security related function shouldn't exist in this repo; it's that I don't think this function provides utility over what's in the standard library, and I think doing things the way they are generally done (via specifying the allow list in the regex) as opposed to what's suggested here is actually the more secure option.

PR #83, opened 2023-08-08
feat(awsutil-v2): implement awsutil for aws-sdk-go-v2

ddebko:
## Summary

This PR is adding a new major version of `awsutil`. The breaking change that requires a major version release is using the latest aws sdk: `aws-sdk-go-v2`. 

### Disclaimer:

Feel free to update to this major release change if you are migrating to `aws-sdk-go-v2`. The Vault team should be cautious of updating to this version due to potential breaking changes from behavior differences between the `aws-sdk-go` libraries.

### Note:

- This PR does not replicate the exact behaviors of the original v1 `awsutil` package. For example, environment variables are always loaded by default in v2. The aws-sdk-go-v2 also doesn't support `RemoteCredProviders`. GetRegion no longer can get a default region value from `ec2metadata` because it does not exist and the replacement `imds` pkg does not support the same feature.
- Custom endpoint resolvers are now associated with a client object (ex: iam, sts) rather than a aws.Config
- withStsEndpoint type was changed from a string to sts.EndpointResolverV2. This option flag was relabeled to withStsEndpointResolver.
- withIamEndpoint type was changed from a string to iam.EndpointResovlerV2. This option flag was relabeled to withIamEndpointResolver.

The benefits of changing the endpoint options from a string to their respective EndpointResolverV2 definition type is that now we can enable complex routing to a set of different endpoints based on fields defined in the resolverV2 struct. Please follow this [link](https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/#endpointresolverv2-parameters) to read more about the EndpointResolverV2 option. Example:

```
type resolverV2 struct {
    // you could inject additional application context here as well
    region string
}

func (*resolverV2) ResolveEndpoint(ctx context.Context, params s3.EndpointParameters) (
        smithyendpoints.Endpoint, error,
    ) {
    if region == "A" {
        return smithyEndpoints.Endpoint{
            URI: url.Parse("https://custom.service.endpoint/"),
        }
    }
    if region == "B" {...}

    // delegate back to the default v2 resolver otherwise
    return s3.NewDefaultEndpointResolverV2().ResolveEndpoint(ctx, params)
}
```


jefferai:
@ddebko rather than making this `awsutilv2` just make this `v2` of the `awsutil` package. That will make it a separate Go module that one must explicitly import and the current version can still be modified if desired. But the general line of development is clearly to catch up with the current, supported version of the AWS SDK.

sgmiller:
Jeff's suggestion might make it easier to review too.

ddebko:
@jefferai @sgmiller I have updated the PR to move the code into a subdirectory/package called v2 in awsutil. I have 2 `TODO` comments for the custom endpoint resolvers and I was hoping to get some feedback on the behavior of the resolvers.

jefferai:
Just a note to please move the files back to the main directory and update `go.mod` before tagging as v2.

Arguably it'd be more useful to have them there now as we could see the differences instead of it all being new code.

awsutil/v2/generate_credentials.go
```diff
@@ -0,0 +1,304 @@
...
+func (c *CredentialsConfig) GenerateCredentialChain(ctx context.Context, opt ...Option) (*aws.Config, error) {
+	opts, err := getOpts(opt...)
+	if err != nil {
+		return nil, fmt.Errorf("error reading options in GenerateCredentialChain: %w", err)
+	}
+
+	// Have one or the other but not both and not neither
+	if (c.AccessKey != "" && c.SecretKey == "") || (c.AccessKey == "" && c.SecretKey != "") {
+		return nil, fmt.Errorf("static AWS client credentials haven't been properly configured (the access key or secret key were provided but not both)")
+	}
```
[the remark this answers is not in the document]
ddebko: If both the AccessKey and SecretKey are empty strings, then the conditional will return false. This conditional check is here to validate that both AccessKey and SecretKey are present. I think that AWS returns a generic credential error message if you don't supply the AccessKey, but provide the SecretKey (and vice-versa). By adding this conditional check, we are providing the user more context into the issue.

sgmiller (approved):
Afaict it looks good to me.

awsutil/v2/generate_credentials.go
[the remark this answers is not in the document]
hugoamvieira: Sorry, I misread the `if` statement 🤦

hugoamvieira (approved):

jefferai (changes requested):
Please make sure to make the necessary changes to `go.mod` given the version bump.

Unless the plan is to mark this as v1.0.0 since the current tags are still in `v0`?

ddebko:
@jefferai what changes do you expect to see in the go.mod file?
I thought we were going to version bump based on release tags as we do with the other repos. Does this still require a bump in the go.mod? Is this the change you would like to see added?
```
module github.com/hashicorp/go-secure-stdlib/awsutil/v1
```

ddebko:
@jefferai what changes do you expect to see in the go.mod file?
I thought we were going to version bump based on release tags as we do with the other repos. Does this still require a bump in the go.mod? Is this the change you would like to see added?
```
module github.com/hashicorp/go-secure-stdlib/awsutil/v1
```

ddebko:
> Please make sure to make the necessary changes to `go.mod` given the version bump.
> 
> Unless the plan is to mark this as v1.0.0 since the current tags are still in `v0`?

I think we should leave the go.mod as is because the next major version bump should be v1.0.0

awsutil/README.md
```diff
@@ -0,0 +1,37 @@
...
+)
+if err != nil {
+    return err
+}
+
+s3Client := s3.NewFromConfig(config)
+
+```
+
+## Contributing to v0
```
austingebauer, line 31: Thanks!

awsutil/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+To push a bug fix or feature for awsutil `v0`, branch out from the `awsutil/v0` branch.
```
austingebauer, line 33: Mind adding a link to the branch? Can go around the `awsutil/v0` text. Doesn't look like the branch exists yet.

ddebko: I created the branch and added a link to this line.

jefferai (approved):
Approving to unblock merge access

benashz:
For some reason the changes from 1a4b95565d57828244cbb7879b8d3a8f3d20522a were not ported over. The PR description does not seem to mention why they were dropped.

PR #87, opened 2023-08-29
Implement Diagnose from runner.Runner

tomhjp:
* Adds Diagnose
* Splits config's image option into image and tag to handle inputs more explicitly
* Other small tidy ups


fairclothjm (approved):

tomhjp:
Thanks!

PR #89, opened 2023-08-31
Set unix socket volume's bind mount propagation to private

tomhjp:
gVisor's runsc only supports "private" and "slave" modes, and we don't need any propagation anyway so just default to private. See here for some good documentation on the setting: https://docs.docker.com/storage/bind-mounts/#configure-bind-propagation

In particular:

> `private`: The mount is private. Sub-mounts within it are not exposed to replica mounts, and sub-mounts of replica mounts are not exposed to the original mount.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -37,6 +37,37 @@ jobs:
     - name: Build
       run: cd ${{ matrix.module }} && go build ./...
 
+    - name: Install gVisor
+      if: ${{ matrix.module == 'plugincontainer' }}
+      run: |
+        (
+          set -e
+          ARCH=$(uname -m)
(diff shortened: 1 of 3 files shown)


.github/workflows/go.yml
```diff
@@ -37,6 +37,37 @@ jobs:
     - name: Build
       run: cd ${{ matrix.module }} && go build ./...
 
+    - name: Install gVisor
```
tomhjp, line 40: This is a lightly modified script from gVisor's installation instructions: https://gvisor.dev/docs/user_guide/install/

swenson (approved):

tomhjp:
Thanks!

PR #90, opened 2023-09-04
plugincontainer: Drop all capabilities from plugin containers

tomhjp:
This makes several breaking API changes, so will necessitate a bump to v0.2.0, but should provide us with a better API in the longer term.

Builds on the additions in https://github.com/hashicorp/go-plugin/pull/277 to enable dropping all default capabilities from the plugin container. We now let go-plugin set the PLUGIN_UNIX_SOCKET_GROUP env var, and override the value it sets for PLUGIN_UNIX_SOCKET_DIR because the plugin's view of that directory is different. As a result, renamed `UnixSocketGroup` as `GroupAdd` to more accurately reflect that it now only controls the container's GroupAdd setting.

As that requires a breaking change, I also updated the API for creating a RunnerFunc to make it a bit cleaner and reduce unnecessarily exported API surface like `ContainerRunner` which was useless anyway when directly created from outside the package - the best place to see the result of this is in container_runner_test.go or example/bidirectional/main.go.

Lastly this PR also removes the container's env from the Diagnose output, as it could reasonably include secrets. I'm planning to re-add that capability in a separate PR but behind a `Debug` configurable.

### DIFF
--- a/plugincontainer/container_runner.go
+++ b/plugincontainer/container_runner.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"path"
 	"runtime"
+	"strconv"
 	"strings"
 
 	"github.com/docker/docker/api/types"
@@ -133,7 +134,7 @@ func (cfg *Config) NewContainerRunner(logger hclog.Logger, cmd *exec.Cmd, hostSo
 	}
 
 	if cfg.GroupAdd != 0 {
-		hostConfig.GroupAdd = append(hostConfig.GroupAdd, fmt.Sprintf("%d", cfg.GroupAdd))
+		hostConfig.GroupAdd = append(hostConfig.GroupAdd, strconv.Itoa(cfg.GroupAdd))
 	}
 
 	// Network config.
--- a/plugincontainer/container_runner.go
+++ b/plugincontainer/container_runner.go
@@ -2,7 +2,6 @@ package plugincontainer
 
 import (
 	"context"
-	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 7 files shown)


plugincontainer/container_runner.go
```diff
@@ -116,6 +112,7 @@ func NewContainerRunner(logger hclog.Logger, cmd *exec.Cmd, cfg *config.Containe
 			Memory:       cfg.Memory,       // Memory limit in bytes.
 			CgroupParent: cfg.CgroupParent, // Parent Cgroup for the container.
 		},
+		CapDrop: []string{"ALL"},
```
tomhjp, line 115: It's kind of small, but this is the most important change in this PR :)

plugincontainer/container_runner.go
```diff
@@ -135,17 +132,16 @@ func NewContainerRunner(logger hclog.Logger, cmd *exec.Cmd, cfg *config.Containe
 		},
 	}
 
-	if cfg.UnixSocketGroup != "" {
-		containerConfig.Env = append(containerConfig.Env, fmt.Sprintf("%s=%s", plugin.EnvUnixSocketGroup, cfg.UnixSocketGroup))
-		hostConfig.GroupAdd = append(hostConfig.GroupAdd, cfg.UnixSocketGroup)
+	if cfg.GroupAdd != 0 {
+		hostConfig.GroupAdd = append(hostConfig.GroupAdd, fmt.Sprintf("%d", cfg.GroupAdd))
```
swenson, line 136: Very minor suggestion: it's slightly faster (and maybe as readable?) to use `strconv.Itoa(x)` instead of `fmt.Sprintf("%d", x)`.

```suggestion
		hostConfig.GroupAdd = append(hostConfig.GroupAdd, strconv.Itoa(cfg.GroupAdd))
```

swenson (approved):
LGTM

plugincontainer/container_runner.go
tomhjp: Oh yeah, good call! Thanks

tomhjp:
Thanks!

PR #91, opened 2023-09-04
plugincontainer: Add debug mode

tomhjp:
Adds a debug mode which does extra work to help diagnose why a plugin may have failed to start, and make it easier for users to re-run their plugins interactively to better understand their behaviour.

In particular, we disable the `AutoRemove` option in the Docker SDK when debug is enabled, otherwise we may end up in a race with the container runtime to salvage debug information before the container is killed and cleaned up.

We also only print environment variables in the diagnose notes if debug is enabled, as they could reasonably contain secrets the plugin needs to consume.

### DIFF
--- a/plugincontainer/container_runner.go
+++ b/plugincontainer/container_runner.go
@@ -2,7 +2,6 @@ package plugincontainer
 
 import (
 	"context"
-	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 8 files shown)


plugincontainer/container_runner.go
```diff
@@ -313,9 +326,92 @@ func (c *containerRunner) Diagnose(ctx context.Context) string {
 	}
 	notes += fmt.Sprintf("GroupAdd: %v\n", c.hostConfig.GroupAdd)
 
+	if c.debug {
+		notes += "Env:\n"
+		const envClientCert = "PLUGIN_CLIENT_CERT"
```
swenson, line 331: Maybe I'm missing something, but is this constant used?

swenson (approved):
LGTM

tomhjp:
Thanks!

PR #93, opened 2023-09-13
plugincontainer: Fix container logs context to have no timeout/cancel

tomhjp:
Found in manual testing of the Vault release 1.15.0-rc1

The context passed to `Start` by go-plugin is cancelled once the plugin has successfully started and communicated its address, but the lifetime of the container log streaming needs to be uncoupled from that. The log streaming will stop when it encounters an EOF error, which will happen when the container is killed.


swenson (approved):
LGTM

tomhjp:
Thanks!

PR #107, opened 2023-10-27
plugincontainer: Support plugins in rootless runtime with nonroot user

tomhjp:
Support running as non-root container users under rootless container runtimes.

* Adds a new configuration option `Config.Rootless`, which should be used when both the runtime and the container user are running as non-root. It sets up default ACLs on the host socket directory, and gives the container the DAC_OVERRIDE capability to ensure each side can write to the shared Unix socket and folder despite each being owned by one side and the other side being a different user on the host.
* Pulls in https://github.com/hashicorp/go-plugin/pull/288 and uses the new `GRPCBrokerMultiplexing` option to eliminate host-side sockets and reduce the `--host-uds` flag from `all` to `create`, which means no Unix domain sockets from the host will ever be available inside gVisor containers.
* Stop skipping rootless + nonroot matrix compatibility tests for gVisor
* Drop `podman` tests for now - they are different enough to be a pain but we don't currently have strong requirements to support podman. It could still get re-added at a later date though.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -87,6 +87,9 @@ jobs:
           }
         }
         EOF
+
+        export PATH="$HOME/bin:$PATH"
+        sudo setcap cap_ipc_lock=ep "$(which rootlesskit)"
         systemctl --user restart docker
     
     - name: Test
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -89,7 +89,6 @@ jobs:
         EOF
 
         export PATH="$HOME/bin:$PATH"
-        sudo setcap cap_ipc_lock=ep "$(which rootlesskit)"
         systemctl --user restart docker
     
     - name: Test
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -58,7 +58,7 @@ jobs:
             "runsc": {
               "path": "/usr/local/bin/runsc",
               "runtimeArgs": [
-                "--host-uds=all"
+                "--host-uds=create"
               ]
             }
           }
@@ -80,7 +80,7 @@ jobs:
             "runsc": {
               "path": "/usr/local/bin/runsc",
               "runtimeArgs": [
-                "--host-uds=all",
+                "--host-uds=create",
                 "--ignore-cgroups"
               ]
             }
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -89,19 +89,6 @@ jobs:
         EOF
         systemctl --user restart docker
     
-    - name: Install rootless podman
-      if: ${{ matrix.module == 'plugincontainer' }}
-      run: |
-        sudo apt-get install -y podman slirp4netns fuse-overlayfs
-        mkdir -p ~/local/bin
-        RUNSC_SCRIPT=~/local/bin/runsc.podman
-        tee "${RUNSC_SCRIPT}" <<EOF
-        #!/bin/bash
-        /usr/local/bin/runsc --host-uds=all --ignore-cgroups "\$@"
-        EOF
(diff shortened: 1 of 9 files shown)


tomhjp:
Note: After some offline feedback, this is still a bit of a work in progress. @mickael-hc pointed out the 0o777 folder could be taken over, so I tried to fix that by ensuring it's inside a more limited folder in ce76700, but for some reason that broke the rootless mlock tests.

tomhjp:
Just pushed some updates that switch from a 0o777 folder to using the DAC_OVERRIDE capability instead. I don't think I'd want to support that for `runc`, because `runc` will require that `rootlesskit` on the host has the capability itself (and it's a very powerful capability), but it seems more reasonable for `runsc` which doesn't need that because AFAIU it has its own userspace implementation instead of relying on the real underlying kernel.

.github/workflows/go.yml
```diff
@@ -58,7 +58,7 @@ jobs:
             "runsc": {
               "path": "/usr/local/bin/runsc",
               "runtimeArgs": [
-                "--host-uds=all"
+                "--host-uds=create"
```
tomhjp, line 61: If we stick with `DAC_OVERRIDE` when `Config.Rootless` is specified, the utility of the new `GRPCBrokerMultiplex` option is greatly reduced because the container would be able to write to host-created sockets without any additional changes. The only remaining benefit is this switch from `--host-uds=all` to `--host-uds=create`. I do think it's nice to know that it's impossible for host-created sockets to become visible and writable in the container, but it's less crucial than it was for the previous implementation.

I need to mull over the complexity vs benefit tradeoffs here a bit more before making a decision, but I'd welcome any other inputs on that.

plugincontainer/compatibility_matrix_test.go
```diff
@@ -40,28 +40,68 @@ func (m matrixInput) String() string {
...
-					})
-				}
-			}
+	testCases := [][]func(matrixInput) matrixInput{
+		{withRootlessEngine(true), withRootlessEngine(false)},
+		{withContainerRuntime(runtimeRunc), withContainerRuntime(runtimeRunsc)},
+		{withRootlessUser(true), withRootlessUser(false)},
+		{withMlock(true), withMlock(false)},
+	}
+	// Run a test for all combinations of 4 binary choices.
```
swenson, line 84: I like this a lot. :)

I thought of a few slightly simpler ways of doing it, but they are minor optimizations at best, and probably not worth doing as they may or may not improve readability. (Like: switching to a Gray code to eliminate the `for`-`j` loop, or manipulating the boolean variables in `matrixInput` directly instead of using the functions.)

I also looked for some sort of Python `itertools.product`-like method that could do this for us, but sadly didn't find anything.

swenson (approved):
LGTM

plugincontainer/compatibility_matrix_test.go
tomhjp: Yeah a permutations/product method would make this pattern much easier to keep if we ever have more than 2 options for one of the inputs. I did make the functions a bit terser with inspiration from this comment, but otherwise agreed about other tweaks, even though they're fun :)

tomhjp:
Thanks! It took me a while to fully convince myself of the need for GroupAdd and Rootless together. For reference, [this](https://www.usenix.org/legacy/events/usenix03/tech/freenix03/full_papers/gruenbacher/gruenbacher.pdf) paper (attached for posterity) had the most thorough explanation I could find for putting the full picture together.

[POSIX Access Control Lists on Linux, Andreas Grunbacher.pdf](https://github.com/hashicorp/go-secure-stdlib/files/13419787/POSIX.Access.Control.Lists.on.Linux.Andreas.Grunbacher.pdf)

PR #109, opened 2023-11-13
Allow compiling to WASM

swenson:
By updating go-sockaddr to 1.0.6 and shuffling some files around so that build tags would ignore irrelevant files for JS/WASM (such as the plugincontainer stuff that doesn't even make sense when compiling to JS/WASM).


swenson:
(Waiting on https://github.com/mitchellh/cli/pull/103 before `listenerutil` will build under JS/WASM.)

swenson:
(Also waiting on https://github.com/hashicorp/raft/pull/577 for `configutil` to build.)

configutil/go.sum
```diff
@@ -56,6 +56,8 @@ github.com/hashicorp/go-secure-stdlib/tlsutil v0.1.3 h1:xbrxd0U9XQW8qL1BAz2XrAjA
 github.com/hashicorp/go-secure-stdlib/tlsutil v0.1.3/go.mod h1:LWq2Sy8UoKKuK4lFuCNWSjJj57MhNNf2zzBWMtkAIX4=
 github.com/hashicorp/go-sockaddr v1.0.5 h1:dvk7TIXCZpmfOlM+9mlcrWmWjw/wlKT+VDq2wMvfPJU=
 github.com/hashicorp/go-sockaddr v1.0.5/go.mod h1:uoUUmtwU7n9Dv3O4SNLeFvg0SxQ3lyjsj6+CCykpaxI=
+github.com/hashicorp/go-sockaddr v1.0.6 h1:RSG8rKU28VTUTvEKghe5gIhIQpv8evvNpnDEyqO4u9I=
```
tomhjp, line 59: Just a guess, but I think this needs a `go mod tidy`, which will likely remove v1.0.5 from go.sum.

listenerutil/go.sum
```diff
@@ -35,6 +35,8 @@ github.com/hashicorp/go-secure-stdlib/tlsutil v0.1.3 h1:xbrxd0U9XQW8qL1BAz2XrAjA
 github.com/hashicorp/go-secure-stdlib/tlsutil v0.1.3/go.mod h1:LWq2Sy8UoKKuK4lFuCNWSjJj57MhNNf2zzBWMtkAIX4=
 github.com/hashicorp/go-sockaddr v1.0.5 h1:dvk7TIXCZpmfOlM+9mlcrWmWjw/wlKT+VDq2wMvfPJU=
 github.com/hashicorp/go-sockaddr v1.0.5/go.mod h1:uoUUmtwU7n9Dv3O4SNLeFvg0SxQ3lyjsj6+CCykpaxI=
+github.com/hashicorp/go-sockaddr v1.0.6 h1:RSG8rKU28VTUTvEKghe5gIhIQpv8evvNpnDEyqO4u9I=
```
tomhjp, line 38: `go mod tidy`

tlsutil/go.sum
```diff
@@ -5,6 +5,8 @@ github.com/hashicorp/go-secure-stdlib/strutil v0.1.2 h1:kes8mmyCpxJsI7FTwtzRqEy9
 github.com/hashicorp/go-secure-stdlib/strutil v0.1.2/go.mod h1:Gou2R9+il93BqX25LAKCLuM+y9U2T4hlwvT1yprcna4=
 github.com/hashicorp/go-sockaddr v1.0.5 h1:dvk7TIXCZpmfOlM+9mlcrWmWjw/wlKT+VDq2wMvfPJU=
 github.com/hashicorp/go-sockaddr v1.0.5/go.mod h1:uoUUmtwU7n9Dv3O4SNLeFvg0SxQ3lyjsj6+CCykpaxI=
+github.com/hashicorp/go-sockaddr v1.0.6 h1:RSG8rKU28VTUTvEKghe5gIhIQpv8evvNpnDEyqO4u9I=
```
tomhjp, line 8: `go mod tidy`

plugincontainer/container_reattach_linux.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/docker/docker/client"
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-plugin/runner"
+)
+
+func ReattachFunc(logger hclog.Logger, id, hostSocketDir string) (runner.AttachedRunner, error) {
+	if runtime.GOOS != "linux" {
+		return nil, ErrUnsupportedOS
+	}
```
tomhjp, line 19: May as well delete this now

```suggestion
```

plugincontainer/container_runner.go
```diff
@@ -4,429 +4,15 @@
...
-	"github.com/hashicorp/go-hclog"
-	"github.com/hashicorp/go-plugin"
-	"github.com/hashicorp/go-plugin/runner"
 )
 
 var (
-	_ runner.Runner = (*containerRunner)(nil)
-
-	errUnsupportedOS = errors.New("plugincontainer currently only supports Linux")
+	ErrUnsupportedOS = errors.New("plugincontainer currently only supports Linux")
```
tomhjp, line 11: Are there any external consumers that need to check for this sentinel error? I'd prefer to keep it unexported until there's a concrete need to export it.

plugincontainer/container_runner.go
```diff
@@ -4,429 +4,15 @@
...
 
 	// ErrSHA256Mismatch is returned when starting a container without any
 	// images available where the provided sha256 matches the image and tag.
 	ErrSHA256Mismatch = errors.New("SHA256 mismatch")
 )
 
 const pluginSocketDir = "/tmp/go-plugin-container"
```
tomhjp, line 18: This can live in the _linux file as well.

plugincontainer/container_runner_linux.go
```diff
@@ -0,0 +1,419 @@
...
+	sha256 string
+	id     string
+	debug  bool
+}
+
+// NewContainerRunner must be passed a cmd that hasn't yet been started.
+func (cfg *Config) NewContainerRunner(logger hclog.Logger, cmd *exec.Cmd, hostSocketDir string) (runner.Runner, error) {
+	if runtime.GOOS != "linux" {
+		return nil, ErrUnsupportedOS
+	}
```
tomhjp, line 55:
```suggestion
```

plugincontainer/container_runner_test.go
```diff
@@ -22,7 +22,7 @@ import (
 func TestNewContainerRunner_config(t *testing.T) {
 	if runtime.GOOS != "linux" {
```
tomhjp, line 23: It would be nice to fully move off `runtime.GOOS` checks and move this `if` statement into its own separate test that's also conditionally compiled for `!linux`.

tomhjp (commented):
LGTM with some minor comments. I avoided the conditional compilation path before because it makes the development experience more painful on non-Linux, but that's a secondary concern.

plugincontainer/container_runner.go
swenson: Not that I know of, but I was more thinking that if we are returning a fixed error, might as well make it easy for people to check it with `errors.Is` if they want to?

tomhjp: Yeah agreed it can be a nice pattern to enable, but in this case it's easy for clients to check whether this will be the cause of the error on their own (granted, with some insider knowledge of the library version they're using). And I'm kind of hoping in the long run we can delete the error altogether by supporting TCP sockets for other OSs.

tomhjp (commented):
I'm not sure how long the upstream `listenerutil` fix is going to take to get fixed, so once `configutil` is fixed, I would be supportive of excluding it from the CI in order to get the rest of this PR merged sooner.

swenson:
All dependencies fixed now :)

Although this moves from `micthellh/cli` to `hashicorp/cli`, this *shouldn't* require users of this library to upgrade immediately: the only usage is with`cli.Ui`, which is a very simple interface that both versions are compatible with.

configutil/go.mod
```diff
@@ -2,9 +2,15 @@ module github.com/hashicorp/go-secure-stdlib/configutil/v2
 
 go 1.20
 
+replace (
+	github.com/hashicorp/go-secure-stdlib/listenerutil => ../listenerutil
+	github.com/hashicorp/go-secure-stdlib/parseutil => ../parseutil
+	github.com/hashicorp/go-secure-stdlib/pluginutil/v2 => ../pluginutil
+)
```
tomhjp, line 9: Are these left in on purpose?

tomhjp (commented):
Nice job on the `hashicorp/cli` changes! Just one question on the `replace` directives

configutil/go.mod
swenson: Definitely on purpose, but I'm not sure if they are strictly necessary after merging. Without them, the build seemed to fail because it kept pulling in the previous tag for those libraries.

tomhjp: Oh I see now, that makes sense. I guess the alternative would be to split this PR into multiple PRs and merge the updates to dependent libraries once the dependencies have been tagged. I think I have a slight preference for that approach because it makes it easier to reason about exactly what version of the code is being consumed, and also has a better defined solution for making breaking changes in modules. That's also the approach that other modules in the repo consistently take, so it would be good to keep that uniformity. WDYT? cc @jefferai as the repo owner

tomhjp: Ah my mistake, you're right about it actually not affecting consumers, from https://go.dev/ref/mod#go-mod-file-replace:

> replace directives only apply in the main module’s go.mod file and are ignored in other modules.

In that case I don't really have a preference, so as it is works for me 👍

tomhjp (approved):
Thanks! Sorry for the merge conflicts from my PR

swenson:
Thanks! No problem on the conflicts :)

PR #117, opened 2024-02-06
return blank cleanup func for plugin.InmemCreationFunc

bosorawis:
Return a blank cleanup func instead of `nil` to prevent accidental panic from `defer cleanup()`

### DIFF
--- a/pluginutil/pluginutil.go
+++ b/pluginutil/pluginutil.go
@@ -168,6 +168,10 @@ func CreatePlugin(plugin *PluginInfo, opt ...Option) (interface{}, func() error,
 	var file fs.File
 	var name string
 
+	cleanup := func() error {
+		return nil
+	}
+
 	switch {
 	case plugin == nil:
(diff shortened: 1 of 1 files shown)


guygrigsby (approved):

jefferai:
This looks good - wherever you saw the panic can you also add a nil check there? Not sure if it was in this package or elsewhere.

bosorawis:
@jefferai I ran into the panic in HCP Boundary code, not sure it also happens elsewhere. This PR is referenced above

jefferai (approved):

jefferai (changes requested):
On line 254 we create a cleanup func that is returned in success scenarios; depending on various conditions we modify that cleanup func later. Rather than create a func inline here, can you move creation of that cleanup above this (but returning nil) and simply reassign it on line 254?

That just keeps the return pattern consistent throughout.

bosorawis:
@jefferai done! thank you!

jefferai:
Tagged v2.0.7.

PR #118, opened 2024-02-06
Fix StrListDelete side effect.

ncabatoff:


jasonodonnell (approved):
Nice find! Go mod seems unrelated?

ncabatoff:
Oh yeah, that's from an earlier attempt using `slices`.  Will remove.

strutil/strutil.go
```diff
@@ -367,6 +367,10 @@ func StrListDelete(s []string, d string) []string {
 
 	for index, element := range s {
 		if element == d {
+			// Using the provided slice as the basis for the return value can
+			// result in confusing behaviour, see https://go.dev/play/p/EAtNw4lLugu
+			c := make([]string, len(s))
+			copy(c, s)
```
tomcf-hcp, line 373: You don't actually use `c` here?
Should the next line actually be `return append(c[:index], c[index+1:]...)`?

tomcf-hcp: And instead of copying `s`, would it be better to build `c` from `s[:index]` and `s[index+1:]`?

tomcf-hcp (changes requested):
Adding these changes to the playground did not change side effects.

strutil/strutil.go
ncabatoff: Haha, whoops, nice catch re `c` not being used.  I thought about being more clever in terms of indexing into `c`, but really I just wanted to fix the immediate problem, didn't want to get sucked into micro-optimization.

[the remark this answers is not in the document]
ncabatoff: Fixed, ty!

averche:
It would also be great to enhance this [test](https://github.com/hashicorp/go-secure-stdlib/blob/b4a7f3e5652cb93b2c3c3f42bb2b2c8ed21641ac/strutil/strutil_test.go#L14) to ensure the original slice is not modified.

tomcf-hcp (approved):

strutil/strutil.go
ncabatoff: I thought I responded to this yesterday, but apparently not.  I fixed the not-using-c issue (and the test I added failed without that).  I decided against the more clever approach you suggested, because I didn't want to pre-optimize, and because if I tried to make it optimal I probably wouldn't get around to fixing the bug before I got pulled off this for something else and then forgot about it.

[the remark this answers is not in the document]
ncabatoff: Fixed, thanks!

ncabatoff:
> It would also be great to enhance this [test](https://github.com/hashicorp/go-secure-stdlib/blob/b4a7f3e5652cb93b2c3c3f42bb2b2c8ed21641ac/strutil/strutil_test.go#L14) to ensure the original slice is not modified.

Done.

strutil/strutil_test.go
```diff
@@ -46,6 +46,13 @@ func TestStrListDelete(t *testing.T) {
 	if StrListContains(output, "item3") {
 		t.Fatal("bad: 'item3' should not have been present")
 	}
+
+	// Test that StrListDelete doesn't modify its input
+	input := []string{"a", "b", "c"}
+	output = StrListDelete(input, "b")
```
averche, line 52: Nit: the `output` variable is unused below, should it be replaced with `_` or is compiler OK with it?

averche (approved):

PR #125, opened 2024-04-12
plugincontainer: fix docker options

kerolasa:
The docker module (provided by The Moby Project) has removed obsoleted function options, and replaced them with the ones from container module.

Reference: https://github.com/moby/moby/commit/13f46948dd57c9530906789de2732ab9dd13d004


debasishbsws (approved):

tvoran:
Thanks, this was done in https://github.com/hashicorp/go-secure-stdlib/pull/126

PR #127, opened 2024-05-10
Update docker to v26.1.2+incompatible

VioletHynes:
This is for https://github.com/hashicorp/vault/pull/26953 -- I want to keep this dependency up to date, and this dependency is currently blocking the upgrade in Vault. It seems like a good idea to update in both places.

This will need a new version, since, as the dependency's name suggests, they're incompatible.


VioletHynes:
Closing this. The test is failing and it seems like this was more work than I'd anticipated. I'll go with just keeping them up to date :)

PR #126, opened 2024-04-26
plugincontainer: upgrade docker/docker package to v26.1.0

odvarkadaniel:
I am currently having issues with `hashicorp/vault` importing `go-secure-stdlib/plugincontainer` which still imports an older version of `docker/docker` package which is causing issues with container types being migrated to a different location.

PR #123, opened 2024-04-04
plugincontainer: Fix rootless tests

tomhjp:
It seems like gVisor fixed a bug where a file's capabilities weren't being properly applied when creating the process: https://github.com/google/gvisor/commit/586c38d70081b13b2ed494cef48e99b93956843e

I haven't fully finished investigating yet, but I think it's the above commit that caused the tests to start failing once it got released and used in CI.

PR #128, opened 2024-07-12
chore(deps): Update aws-sdk-go-v2

hugoghx:
Important: Due to underlying changes in the SDK, awsutil/v2 will no
longer automatically load the "default" AWS shared profile if none is
provided. To load a shared profile, you'll need to set the AWS_PROFILE
env var.

---

After merging this PR, a new `awsutil/v2.1.0` tag and GitHub release will be created.


ddebko (approved):
The plugin container job is failing, you might just need to rerun it?
The error looks strange:
```
TestExamplePlugin/runsc/simulated_plugin_error: plugin shut down with non-0 exit code
```

hugoghx:
> The plugin container job is failing, you might just need to rerun it? The error looks strange:
> 
> ```
> TestExamplePlugin/runsc/simulated_plugin_error: plugin shut down with non-0 exit code
> ```

Yeah, I re-ran it once and it just failed again. Currently looking into this - I think it's unlikely that these changes are making it fail though

hugoghx:
Adding @jefferai as we need an authorised user to approve these changes

hugoghx:
I wasn't able to find much on why `plugincontainer` is failing. I re-ran it once more and it failed again - My only guess here is that something changed in the dependencies we use to run that test, given that we always pull `latest` for some of them.

I've also noticed, looking at all the [actions](https://github.com/hashicorp/go-secure-stdlib/actions), that this is the case for other PRs, and the only `plugincontainer` test that has passed was 4 months ago with the last merge to `main`. 

There's also some PRs targeting `plugincontainer` https://github.com/hashicorp/go-secure-stdlib/pull/126 https://github.com/hashicorp/go-secure-stdlib/pull/125 https://github.com/hashicorp/go-secure-stdlib/pull/123

johanbrandhorst (approved):

louisruch:
Oh I see @hugoghx tried to do this a while ago, I am closing this in favor of #143

PR #130, opened 2024-08-02
Add a helper RoundTripper for ignoring unhandleable critical certificate extensions during TLS negotiation.

sgmiller:
A customer has an intermediate certificate outside their control which
contains two RFC 5280 extensions not supported by Go, which causes a
rejection of the connection.  This helper RoundTripper allows us to ignore a
specific list of extensions while verifying the rest of the certificate.


httputil/cert_ext_tripper.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+type ignoreExtensionsRoundTripper struct {
+	base         *http.Transport
+	extsToIgnore []asn1.ObjectIdentifier
+	logger       hclog.Logger
+}
+
+// NewIgnoreUnhandledExtensionsRoundTripper creates a RoundTripper that may be used in an HTTP client which will
+// ignore the provided extensions if present unhandled on a certificate.  If base is nil, the default RoundTripper is used.
```
jefferai, line 25:
```suggestion
// ignore the provided extensions if presently unhandled on a certificate.  If base is nil, the default RoundTripper is used.
```

jefferai, line 25:
```suggestion
// ignore the provided extensions if present unhandled on a certificate.  If base is nil, `http.DefaultClient` is used.
```

httputil/cert_ext_tripper.go
```diff
@@ -0,0 +1,145 @@
...
+func NewIgnoreUnhandledExtensionsRoundTripper(logger hclog.Logger, base http.RoundTripper, extsToIgnore []asn1.ObjectIdentifier) http.RoundTripper {
```
jefferai, line 26: I don't love requiring a logger here as not all users of this library are necessarily using hclog (Boundary, for instance). Looking further down, it's used in one place purely for debugging, and in another to warn with the extension info when the verification is about to error out anyways. If the verification already tells you which extension is unhandled when it errors out (I'm not sure, as tls errors are fairly opaque in Go), I think we should just remove it entirely.

httputil/cert_ext_tripper.go
```diff
@@ -0,0 +1,145 @@
...
+	if len(extsToIgnore) == 0 {
```
jefferai, line 27: Shouldn't this error instead? It seems like it should be an error to be using this lib when you aren't ignoring anything.

httputil/cert_ext_tripper.go
```diff
@@ -0,0 +1,145 @@
...
+		return base
+	}
+	if base == nil {
```
jefferai, line 30: I'm not sure it's safe to be modifying the shared default transport, e.g. if this is called multiple times. We may want to require a transport to be provided, and in fact may want to validate that it _isn't_ `http.DefaultTransport`.

httputil/cert_ext_tripper.go
```diff
@@ -0,0 +1,145 @@
...
+		base = http.DefaultTransport
+	}
+
```
jefferai, line 33: This function is missing a check that the transport's TLSClientConfig is not nil, and erroring if it is. Right now it appears that if a non-TLS transport is provided, including using the DefaultTransport if it has been given no TLS info, you'll panic in RoundTrip.

httputil/cert_ext_tripper.go
```diff
@@ -0,0 +1,145 @@
...
+		// We don't know how to deal with this object, bail
+		return base
+	}
+
+	return &ignoreExtensionsRoundTripper{base: tp, logger: logger, extsToIgnore: extsToIgnore}
+}
+
+func (i *ignoreExtensionsRoundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
+	var domain string
+	if strings.ContainsRune(request.URL.Host, ':') {
```
jefferai, line 45: I believe this will fail for IPv6 since it's valid to have an IP address as the host.

You should do a `net.SplitHostPort` and check the returned error, if any, something like:

```
			host, _, err = net.SplitHostPort(ln.Config.Address)
			if err != nil {
				if strings.Contains(err.Error(), "missing port") {
					host = ln.Config.Address
				} else {
					return fmt.Errorf("error splitting host/port: %w", err)
				}
			}
```

httputil/cert_ext_tripper.go
```diff
@@ -0,0 +1,145 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("could not parse domain from URL host %s", request.URL.Host)
+		}
+	} else {
+		domain = request.URL.Host
+	}
+
+	// Only update our values if the end-user hasn't overridden anything we wanted to do.
+	var perReqTransport *http.Transport
+	if !i.base.TLSClientConfig.InsecureSkipVerify && i.base.TLSClientConfig.VerifyConnection == nil {
```
jefferai, line 57: Missing a nil check on `TLSClientConfig`, which should probably be checked in the constructor as I noted above.

jefferai, line 57: Checking for these could be done in the constructor as they aren't going to change from call to call, and with an appropriate error.

httputil/cert_ext_tripper.go
```diff
@@ -0,0 +1,145 @@
...
+		perReqTransport = i.base.Clone()
+		var tlsConfig *tls.Config
+		if perReqTransport.TLSClientConfig == nil {
```
jefferai, line 60: Here's the nil check, but on the clone :-D

httputil/cert_ext_tripper.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return perReqTransport.RoundTrip(request)
+}
+
+func (i *ignoreExtensionsRoundTripper) customVerifyConnection(tc *tls.Config) func(tls.ConnectionState) error {
+	return func(cs tls.ConnectionState) error {
+		certs := cs.PeerCertificates
+
+		serverName := cs.ServerName
+		if cs.ServerName == "" {
```
jefferai, line 85: From the RFC: `Currently, the only server names supported are DNS hostnames`

This lib will fail if connecting to an IP address since SNI won't be set, even though it's still valid to want to remove unknown extensions.

httputil/cert_ext_tripper.go
```diff
@@ -0,0 +1,145 @@
...
+	var domain string
+	if strings.ContainsRune(request.URL.Host, ':') {
+		var err error
+		domain, _, err = net.SplitHostPort(request.URL.Host)
+		if err != nil {
+			return nil, fmt.Errorf("could not parse domain from URL host %s", request.URL.Host)
+		}
+	} else {
+		domain = request.URL.Host
+	}
```
jefferai, line 53: Note at the end of this block it's still valid for the domain to be an IP address, which will cause problems later (but should be supported).

httputil/cert_ext_tripper.go
sgmiller: Well, it's nilable and isn't used if not provided, but I'm not opposed to leaving it out.

httputil/cert_ext_tripper.go
sgmiller: Yeah, I think that's probably cleaner.

httputil/cert_ext_tripper.go
sgmiller: We don't modify the passed in transport, we just keep it around for cloning or using unmodified directly.

httputil/cert_ext_tripper.go
sgmiller: That happens later, we initialize a TLSConfig if missing in RoundTrip.  This is to avoid needing to modify the base RoundTripper.

httputil/cert_ext_tripper.go
sgmiller: Good call.

httputil/cert_ext_tripper.go
sgmiller: Don't want to modify base in case it's being used in other clients, but you're right about it being nil here.

sgmiller: Maybe the thing to do is clone base in the constructor as well, then we can move that logic to it.

httputil/cert_ext_tripper.go
jefferai: Ah right, I missed the base clone, thanks.

httputil/cert_ext_tripper.go
jefferai: Ah right, I missed the base clone, thanks.

httputil/cert_ext_tripper.go
jefferai: Ah, sorry, I mixed up that if RoundTrip is missing `http.DefaultTransport` will use a system TLS config. 👍

httputil/cert_ext_tripper.go
```diff
@@ -0,0 +1,134 @@
...
+		certs := cs.PeerCertificates
+
+		serverName := cs.ServerName
+		if cs.ServerName == "" && !ipBased {
+			if tc.ServerName == "" {
+				return fmt.Errorf("the ServerName in TLSClientConfig is required to be set when UnhandledExtensionsToIgnore has values")
+			}
+			serverName = tc.ServerName
+		} else if cs.ServerName != tc.ServerName {
+			return fmt.Errorf("x509: connection state server name (%s) does not match requested (%s)", cs.ServerName, tc.ServerName)
```
jefferai, line 89: The `x509` in the front might give someone the impression that it's an error from the standard library, and is not how the error just above is formatted (or any other error).

httputil/cert_ext_tripper_test.go
```diff
@@ -0,0 +1,188 @@
...
+	srvWith := newTLSServer(t, true, "localhost")
+	defer srvWith.Close()
+	// Only works with a host entry
+	// srvWithout := newTLSServer(t, false, "example.com")
+	// defer srvWithout.Close()
+	srvIP := newTLSServer(t, true, "127.0.0.1")
+	defer srvIP.Close()
+
+	tests := []struct {
+		name         string
```
jefferai, line 39: There are no tests for IP (v4/v6) or port handling.

jefferai (approved):

PR #137, opened 2024-10-11
Add test for overflow during multiplication in TestParseDurationSeconds.

sgmiller:
Error out rather than producing negative values when this happens.  Addresses https://hashicorp.atlassian.net/browse/VAULT-5437

### DIFF
--- a/parseutil/parseutil.go
+++ b/parseutil/parseutil.go
@@ -18,7 +18,10 @@ import (
 	"github.com/mitchellh/mapstructure"
 )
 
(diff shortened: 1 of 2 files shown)


jefferai (approved):

parseutil/parseutil.go
```diff
@@ -114,7 +117,7 @@ func ParseDurationSecond(in interface{}) (time.Duration, error) {
 		}
 
 		if v, err := strconv.ParseInt(inp, 10, 64); err == nil {
-			return time.Duration(v) * time.Second, nil
+			return overflowMul(time.Duration(v), time.Second)
```
stevendpclark, line 120: Shouldn't the other multiplications below be modified to leverage the function?

stevendpclark (approved):

PR #141, opened 2024-11-15
Add cryptoutil/rsa, RSA key generation using a seeded DRBG

sgmiller:
In Vault with entropy augmentation, we may be sourcing randomness from a
particularly slow HSM.  RSA needs a lot of randomness to choose primes, and
some customers have seen multiple minute key generation times as a result.

This PR implements RSA key generation where a NIST HMAC DRBG is seeded
from the random source, which is then used to sample unpredictable primes,
resulting in a 5000 - 20000x increase in performance on these slow HSMs.

Additionally, these functions test whether the underlying random source
is crypto/rand, and if so, skips this process and behaves as before.

### DIFF
--- a/cryptoutil/go.mod
+++ b/cryptoutil/go.mod
@@ -1,0 +1,11 @@
+module github.com/hashicorp/go-secure-stdlib/rsa
+
+go 1.23.0
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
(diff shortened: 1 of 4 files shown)


stevendpclark (commented):
Overall looks good to me, some small nits

stevendpclark (approved):
👍

PR #153, opened 2025-01-09
permitpool: add new module

johanbrandhorst:
The permitpool module contains a single package that implements a pool of permits, allowing limited number of concurrent executions.

### DIFF
--- a/permitpool/permitpool.go
+++ b/permitpool/permitpool.go
@@ -1,0 +1,49 @@
+// Package permitpool exposes a synchronization primitive for
+// limiting the number of concurrent operations. See the Pool
+// example for a simple use case.
+package permitpool
+
+import "context"
+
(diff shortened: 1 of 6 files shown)


permitpool/permitpool.go
```diff
@@ -0,0 +1,38 @@
+package permitpool
```
tmessi, line 1: Can you add a package doc to describe this package? It might also be helpful to provide an example or two of how this might be used, either in the package doc, or with run-able examples.

jefferai (dismissed):
LGTM but please add package doc as @tmessi suggested

permitpool/permitpool.go
johanbrandhorst: Added a package doc and a runnable example, thanks!

jefferai (approved):

PR #154, opened 2025-01-13
Add backwards compatible changes to ParsePath for extra behaviors

sgmiller:
go-kms-wrapping wants to use ParsePath to supply sensitive values from out
of band means.  But there are three differences that could cause the
inability to supply a value exactly as stored in a file or ENV var with the
existing mechanism or debug misconfigured uses.

1. Trims all values at the moment, and leading/trailing spaces
could be significant.
2. Silently provides "" for an env var that's missing, making it
difficult to debug mistaken env var names or unset env vars.
3. (unlikely, but possible), it's not possible to actually supply a value
that has one of the URL prefixes.

This PR adds an options pattern to solve #1 and #2 by being able to disable
trimming and enable error on missing env var.  The status quo is unchanged.

Adds a "string://" url scheme that returns everything following that prefix
untouched, so you can write a value that contains a URL prefix supported by
ParsePath.  This does have the limitation of not supporting leading/trailing
spaces in the value after string://, as that's invalid in the URL schema.

### DIFF
--- a/listenerutil/parse.go
+++ b/listenerutil/parse.go
@@ -440,14 +440,14 @@ const defaultUiContentSecurityPolicyHeader = "default-src 'none'; script-src 'se
 // Header names consts
 const contentSecurityPolicy = "Content-Security-Policy"
 const strictTransportSecurity = "Strict-Transport-Security"
-const xContentTypeOptions = "X-Content-Type-Options"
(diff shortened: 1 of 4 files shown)


stevendpclark (approved):
👍

jefferai (approved):

PR #156, opened 2025-02-19
[VAULT-21456] Add a drop in replacement for the Go regexp package that interns regular expressions

kubawi:
This PR adds a `regexp` module, which interns regular expressions to avoid excessive memory use.


regexp/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/hashicorp/go-secure-stdlib/regexp
+
+go 1.23
```
bosouza, line 3: didn't the `weak` package come out in 1.24? I guess that means we'll have to upgrade the Go version in order to use this?

regexp/regexp.go
```diff
@@ -0,0 +1,110 @@
...
+	return mustCompile(pattern, regexp.MustCompilePOSIX, posixWeakMap)
+}
+
+// compile handles compiling and interning regular expressions. If the regexp is
+// already interned, a pointer to it is returned from the map. If the regexp is
+// not interned or is nil (since it's a weak pointer), it is compiled and stored
+// in the maps. The regexp is stored in the maps as a weak pointer, so that it
+// can be garbage collected.
+func compile(pattern string, compileFunc func(string) (*regexp.Regexp, error), weakMap map[string]weak.Pointer[regexp.Regexp]) (*regexp.Regexp, error) {
+	l.Lock()
```
bosouza, line 52: This single-lock approach can create a lot of unnecessary contention if a bunch of disjoint compilations are triggered simultaneously right? I wonder if we should consider something like [singleflight](https://pkg.go.dev/github.com/golang/groupcache/singleflight) to allow multiple compilations of different patterns to happen at the same time.

regexp/regexp.go
```diff
@@ -0,0 +1,110 @@
...
+// is the same *regexp.Regexp.  This avoids the compilation cost but more importantly the memory usage.
+//
+// Regular expressions produced from this package are backed by a form of weak-valued map, upon a regexp becoming
+// unreachable, they will be eventually removed from the map and memory reclaimed.
+
+var (
+	weakMap      = make(map[string]weak.Pointer[regexp.Regexp])
+	posixWeakMap = make(map[string]weak.Pointer[regexp.Regexp])
+	reverseMap   = make(map[weak.Pointer[regexp.Regexp]]string)
+	l            sync.RWMutex
```
bosouza, line 20: can't this be a normal `Mutex`?

regexp/regexp_test.go
```diff
@@ -0,0 +1,125 @@
...
+			// to not flake a single time in over two million runs, so it should be good enough.
+			// A single call to runtime.GC() was flaky very frequently in local testing.
+			runtime.GC()
+			runtime.GC()
+
+			// Ensure that the cleanup happened and the maps used for interning regexp are empty
+			l.Lock()
+			require.Len(t, weakMap, 0)
+			require.Len(t, reverseMap, 0)
+			l.Unlock()
```
bosouza, line 80: we could execute these steps in a loop (maybe with a sleep in-between?) until the background thing occurs or a reasonably high timeout is hit, that would guarantee it never flakes.

regexp/regexp_test.go
bosouza, line 1: maybe it would also be cool to have a test that ensures patterns compiled with posix vs non-posix don't conflict with each others lifecycle in the maps.

bosouza (commented):
This looks great, the `weak` package fits like a glove! I'm so glad we could avoid unsafe/reflection :D

regexp/regexp.go
```diff
@@ -0,0 +1,110 @@
...
+func mustCompile(pattern string, compileFunc func(string) *regexp.Regexp, weakMap map[string]weak.Pointer[regexp.Regexp]) *regexp.Regexp {
+	l.Lock()
+	defer l.Unlock()
+	if itemPtr, ok := weakMap[pattern]; ok {
+		ptr := itemPtr.Value()
+		if ptr != nil {
+			return itemPtr.Value()
+		}
+		delete(weakMap, pattern)
+		delete(posixWeakMap, pattern)
```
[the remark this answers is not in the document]
kubawi: Also, if we want to avoid accessing the maps that sit outside the functions, we could do it at the expense of passing the `reverseMap` into both `compile()` and `mustCompile()`, but I am not sure that feels that much cleaner 🤔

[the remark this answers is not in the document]
bosouza: I think I was mostly trying to get at (2), I think it would be clearer to only delete from the map passed in and `reverseMap`, and I see no issue with accessing the global var directly. I also think deleting from both posix and non-posix weak maps causes issues when the same pattern is compiled with the posix and non-posix functions, although that might not be a valid use case. Also I think that use case might be broken anyways because there is a single reverse map for both posix and non-posix.

I do see a scope issue (1) as well, note how in both compile functions the `weakMap` argument shadows the global var of the same name and that argument can be a reference to `posixWeakMap`, in which case
```
delete(weakMap, pattern)
delete(posixWeakMap, pattern)
```
is actually the same as
```
delete(posixWeakMap, pattern)
delete(posixWeakMap, pattern)
```
which is not actually a problem since we could just be deleting from the `weakMap` argument as per the (2) discussion, but it's a bit confusing.

regexp/regexp.go
bosouza: yeah a whole other dependency might be a bit much, but I think we could address it in other ways. For instance, we could compile the regext without holding the mutex, then later lock it and check if another goroutine has compiled the same pattern in which case we return the existing compilation, otherwise we add it to the map. That would allow multiple disjoint compilations to not contend on the lock since the most time consuming step is removed from the critical section. The drawback is that in case of parallel compilations of the same patterns we'd be wasting CPU, but I think that is fine since the main goal here is to not waste RAM which would be accomplished by discarding the duplicate compilation.

regexp/go.mod
kubawi: Fixed in 199d4648724bac134862d050023d2f623ddef39e

regexp/regexp.go
```diff
@@ -0,0 +1,110 @@
+package regexp
+
+import (
+	"regexp"
+	"runtime"
+	"sync"
+	"weak"
+)
+
+// "Interns" compilation of Regular Expressions.  If two regexs with the same pattern are compiled, the result
```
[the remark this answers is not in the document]
kubawi: Updated the comment in 7eaf9deae048396f36ae8edda05805dda6d0aa73

regexp/regexp.go
```diff
@@ -0,0 +1,110 @@
...
+// not interned or is nil (since it's a weak pointer), it is compiled and stored
+// in the maps. The regexp is stored in the maps as a weak pointer, so that it
+// can be garbage collected.
+func mustCompile(pattern string, compileFunc func(string) *regexp.Regexp, weakMap map[string]weak.Pointer[regexp.Regexp]) *regexp.Regexp {
+	l.Lock()
+	defer l.Unlock()
+	if itemPtr, ok := weakMap[pattern]; ok {
+		ptr := itemPtr.Value()
+		if ptr != nil {
+			return itemPtr.Value()
```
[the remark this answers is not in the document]
kubawi: Fixed in 7eaf9deae048396f36ae8edda05805dda6d0aa73

regexp/regexp.go
kubawi: Fixed in 071bc6fb408a627c06f061f6288efeafedae4b99

regexp/regexp.go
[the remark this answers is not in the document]
bosouza: oh I see this time I found a reference to it https://en.wikipedia.org/wiki/String_interning

regexp/regexp.go
```diff
@@ -0,0 +1,110 @@
...
+		delete(posixWeakMap, pattern)
+		delete(reverseMap, itemPtr)
+	}
+	r := compileFunc(pattern)
+	weakPointer := weak.Make(r)
+	weakMap[pattern] = weakPointer
+	reverseMap[weakPointer] = pattern
+	runtime.AddCleanup(r, cleanup, weakPointer)
+	return r
+}
```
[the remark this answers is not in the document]
bosouza: I don't know, that's the core logic of the package that's being duplicated in order to avoid a closure to match the signature with the alternative error handling flavor... that doesn't sit right with me, but if you or other folks really like it better then its fine :sweat_smile:

regexp/regexp.go
raskchanky: Technically we only really need the lock around the map checking/updating, but my hunch is that it doesn't matter either way. I doubt leaving it the way it is will introduce enough lock contention that it will result in a performance hit we can notice, and similarly I doubt that moving the lock to the critical section around map checking/writing will result in a performance increase that we can notice.

regexp/regexp.go
```diff
@@ -0,0 +1,98 @@
...
+func compile(pattern string, compileFunc func(string) (*regexp.Regexp, error), m map[string]weak.Pointer[regexp.Regexp]) (*regexp.Regexp, error) {
+	l.Lock()
+	defer l.Unlock()
+	if itemPtr, ok := m[pattern]; ok {
+		ptr := itemPtr.Value()
+		if ptr != nil {
+			return ptr, nil
+		}
+		delete(m, pattern)
+		delete(posixWeakMap, pattern)
```
raskchanky, line 60: I'm curious about the pattern here where we only insert into whatever map we're called with, `m`, but we delete from both `m` and `posixWeakMap`. It seems like there's a subtle behavior change here, where when `m` is `weakMap`, we're deleting from both maps, but when `m` is `posixWeakMap`, we're only deleting from `posixWeakMap` and not from `weakMap`. I wonder if it would be better to just delete from `m`?

regexp/regexp_test.go
```diff
@@ -0,0 +1,125 @@
+package regexp
+
+import (
+	"github.com/stretchr/testify/require"
+	"regexp"
+	"runtime"
+	"strconv"
+	"sync"
+	"testing"
```
raskchanky, line 9: Can you separate the imports into two groups, like we normally do?

raskchanky (commented):
Generally looks great! I left a few comments.

As I was reading through this, I thought maybe this was a good use case for https://pkg.go.dev/unique which was introduced in go 1.23. On the surface, it seems like exactly what the unique package was made for. The principal downside to that approach is the pointers aren't weak, so you would end up with a map of regexes that sticks around indefinitely.

regexp/regexp_test.go
kubawi: 🤦 I really need to get my IDE to do it right for me. Fixed in 9c844f2ea1502369964e908ca7af8c4e476aeeac

regexp/regexp.go
kubawi: Sorry for confusion, you caught me mid-refactor here and I should have toggled the PR to draft or left a comment.

You are spot on here, Bruno also pointed out this issue in a thread above and we think `cleanup()` might also need to be changed, but I didn't address these issues yet. I will tag you in the other thread, so that we can carry on the conversation there.

regexp/regexp.go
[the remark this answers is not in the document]
kubawi: @raskchanky tagging you in this thread regarding deleting items from maps. It's a mess at the moment, I will try to fix it up tomorrow and comment here after 🫡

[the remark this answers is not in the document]
kubawi: @raskchanky 
> I'm curious about the pattern here where we only insert into whatever map we're called with, m, but we delete from both m and posixWeakMap. It seems like there's a subtle behavior change here, where when m is weakMap, we're deleting from both maps, but when m is posixWeakMap, we're only deleting from posixWeakMap and not from weakMap. I wonder if it would be better to just delete from m?

@bosouza 
> I do see a scope issue (1) as well, note how in both compile functions the weakMap argument shadows the global var of the same name and that argument can be a reference to posixWeakMap, in which case (...)

Thanks both! Fixed in b21b980129c6857a1e6e6fe782790cf53212abf0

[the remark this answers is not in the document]
kubawi: > I also think deleting from both posix and non-posix weak maps causes issues when the same pattern is compiled with the posix and non-posix functions

This could also affect the cleanup function, so I refactored it in ea0066a56d06888a2027f22e7dc22ef0948ba047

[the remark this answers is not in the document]
kubawi: > Also I think that use case might be broken anyways because there is a single reverse map for both posix and non-posix.

That map is using pointers as keys, which should be unique between POSIX and non-POSIX, even if the same string pattern was compiled. I think it's okay as is, but do let me know if I am missing something.

Also, thanks both for this feedback, I am pretty sure that would be a bug if missed. This further hints that the tests still aren't sufficiently good, like Bruno pointed out below. I will spend some time next week thinking about a test that uses both POSIX and non-POSIX regexs and ensures they don't affect each other's lifecycles.

raskchanky (approved):
Looks great!

regexp/regexp.go
kubawi: Fixed in b21b980

regexp/regexp.go
kubawi: I am going to resolve this thread for now. If we ever observe any issues related to lock contention, we can come back and address it.

regexp/regexp_test.go
kubawi: Added in 0b4566a3139256d37ae554f6e66dd89e395b4de5

regexp/regexp_test.go
kubawi: Added an extra test in 5d9a94a24fed4c7de9949f0636623d7956507892

sgmiller (approved):
Looks great!

PR #157, opened 2025-02-25
Creates the NormalizeAddr Func to Format IPv6 Addresses

kheina:
NormalizeAddr takes an address as a string and returns a normalized copy. If the address is a URL, IP Address, or host:port address that includes an IPv6 address, the normalized copy will be conformant with [RFC-5952 §4](https://www.rfc-editor.org/rfc/rfc5952#section-4). If the address cannot be parsed, an error will be returned.

There are two valid formats:
- hosts: `host`
  - may be any of: IPv6 literal, IPv4 literal, dns name, or [sub]domain name
  - IPv6 literals are not required to be encapsulated within square brackets
    in this format
- URIs: `[scheme://] [user@] host [:port] [/path] [?query] [#frag]`
  - format should conform with [RFC-3986 §3](https://www.rfc-editor.org/rfc/rfc3986#section-3) or else the returned address may be parsed and formatted incorrectly
  - hosts containing IPv6 literals MUST be encapsulated within square brackets, as defined in [RFC-3986 §3.2.2](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2) and [RFC-5952 §6](https://www.rfc-editor.org/rfc/rfc5952#section-6)
  - all non-host components are optional

See:
  - [RFC-5952](https://www.rfc-editor.org/rfc/rfc5952)
  - [RFC-3986](https://www.rfc-editor.org/rfc/rfc3986)

### DIFF
--- a/configutil/normalize.go
+++ b/configutil/normalize.go
@@ -1,0 +1,94 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package configutil
+
+import (
+	"fmt"
+	"net"
+	"net/url"
+	"strings"
+)
+
+// general delimiters as defined in RFC 3986
+// See: https://www.rfc-editor.org/rfc/rfc3986#section-2.2
+const genDelims = ":/?#[]@"
+
+func normalizeHostPort(host string, port string, url bool) (string, error) {
+	// fmt.Println("host:", host, "port:", port)
+	if host == "" {
+		return "", nil
+	}
+	if ip := net.ParseIP(host); ip != nil {
+		if url && ip.To4() == nil && port == "" {
+			// this is a unique case, host is ipv6 and requires brackets due to
+			// being part of a url, but they won't be added by net.JoinHostPort
+			// as there is no port
+			return "[" + ip.String() + "]", nil
+		}
+		host = ip.String()
+	} else if strings.Contains(host, ":") {
+		// host is an invalid ipv6 literal.
+		// hosts cannot contain certain reserved characters, including ":"
+		// See: https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2,
+		//      https://www.rfc-editor.org/rfc/rfc3986#section-2.2
+		return "", fmt.Errorf("host contains an invalid IPv6 literal")
+	}
+	if port == "" {
+		return host, nil
+	}
+	return net.JoinHostPort(host, port), nil
+}
+
+// NormalizeAddr takes an address as a string and returns a normalized copy.
+// If the addr is a URL, IP Address, or host:port address that includes an IPv6
+// address, the normalized copy will be conformant with RFC-5942 §4
+//
+// Valid formats include:
+//   - host
+//   - host:port
+//   - scheme://user@host/path?query#frag
+//
+// Note: URLs and URIs must conform with https://www.rfc-editor.org/rfc/rfc3986#section-3
+// or else the returned address may have been parsed and formatted incorrectly
+//
+// See: https://rfc-editor.org/rfc/rfc5952.html
+func NormalizeAddr(address string) (string, error) {
+	if address == "" {
+		return "", fmt.Errorf("empty or invalid hostname")
+	}
+
+	if ip := net.ParseIP(address); ip != nil {
+		return ip.String(), nil
+	}
+
+	if strings.HasPrefix(address, "[") && strings.HasSuffix(address, "]") {
+		return NormalizeAddr(address[1 : len(address)-1])
+	}
+
+	if host, port, err := net.SplitHostPort(address); err == nil {
+		return normalizeHostPort(host, port, false)
+	}
+
+	if u, err := url.ParseRequestURI(address); err == nil {
+		if u.Host, err = normalizeHostPort(u.Hostname(), u.Port(), true); err != nil {
+			return "", err
+		}
+		return u.String(), nil
+	}
+	// if the provided address does not have a scheme provided, attempt to
+	// provide one and re-parse the result. this is done by looking for the
+	// first general delimiter and checking if it exists or if it's not a colon
+	// See: https://www.rfc-editor.org/rfc/rfc3986#section-3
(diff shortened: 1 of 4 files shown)


configutil/normalize.go
```diff
@@ -0,0 +1,94 @@
...
+	"net/url"
+	"strings"
+)
+
+// general delimiters as defined in RFC 3986
+// See: https://www.rfc-editor.org/rfc/rfc3986#section-2.2
+const genDelims = ":/?#[]@"
+
+func normalizeHostPort(host string, port string, url bool) (string, error) {
+	// fmt.Println("host:", host, "port:", port)
```
hugoghx, line 18: Debug line

```suggestion
```

configutil/normalize.go
```diff
@@ -0,0 +1,94 @@
...
+//   - host:port
+//   - scheme://user@host/path?query#frag
+//
+// Note: URLs and URIs must conform with https://www.rfc-editor.org/rfc/rfc3986#section-3
+// or else the returned address may have been parsed and formatted incorrectly
+//
+// See: https://rfc-editor.org/rfc/rfc5952.html
+func NormalizeAddr(address string) (string, error) {
+	if address == "" {
+		return "", fmt.Errorf("empty or invalid hostname")
```
hugoghx, line 58: This case is literally just empty input, so maybe the error can be a little more precise:
```suggestion
		return "", fmt.Errorf("empty address")
```

configutil/normalize.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if strings.HasPrefix(address, "[") && strings.HasSuffix(address, "]") {
+		return NormalizeAddr(address[1 : len(address)-1])
+	}
+
+	if host, port, err := net.SplitHostPort(address); err == nil {
+		return normalizeHostPort(host, port, false)
+	}
+
+	if u, err := url.ParseRequestURI(address); err == nil {
```
hugoghx, line 73: [url.ParseRequestURI](https://pkg.go.dev/net/url#ParseRequestURI) assumes the input has no `#fragment` suffix. [Testing what happens](https://go.dev/play/p/1Y5vl9M9XlT) when you throw it in there anyway, we get an error.

Following this function, it seems like any URL with a fragment suffix would fall in the `return "", fmt.Errorf("unable to normalize given address")` case.

At this stage, I don't know if this is a problem, but just wanted to raise this up.

hugoghx (commented):
Just a first pass 👀

configutil/normalize.go
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
```
louisruch, line 2: This should be MPL-2.0

configutil/normalize_test.go
```diff
@@ -0,0 +1,534 @@
...
+			name:     "ipv6 URL RFC-5952 4.3 conformance downcase hex letters",
+			address:  "https://[2001:DB8:AC3:FE4::1]:8200",
+			expected: "https://[2001:db8:ac3:fe4::1]:8200",
+		},
+		{
+			name:     "ipv6 destination address with port RFC-5952 4.3 conformance downcase hex letters",
+			address:  "username@[2001:DB8:AC3:FE4::1]:8200",
+			expected: "username@[2001:db8:ac3:fe4::1]:8200",
+		},
+		// NOTE: these tests are NOT conformant with the URI spec https://www.rfc-editor.org/rfc/rfc3986#section-3
```
louisruch, line 474: I assume these are in here for sake of review in comparing to the tests done in the Vault package and you will eventually take them out?

configutil/normalize_test.go
```diff
@@ -0,0 +1,534 @@
+package configutil
```
louisruch, line 1: We should add the license header here as well

configutil/normalize_test.go
kheina: yes

configutil/normalize.go
```diff
@@ -0,0 +1,106 @@
...
+// general delimiters as defined in RFC-3986 §2.2
+// See: https://www.rfc-editor.org/rfc/rfc3986#section-2.2
+const genDelims = ":/?#[]@"
+
+func normalizeHostPort(host string, port string, url bool) (string, error) {
+	if host == "" {
+		return "", fmt.Errorf("empty hostname")
+	}
+	if ip := net.ParseIP(host); ip != nil {
+		if url && ip.To4() == nil && port == "" {
```
jefferai, line 22: Might be nice to actually check `To16` because if somehow it parses but neither To4 or To16 work then something is wrong, but it will make it very clear that yes, it is absolutely an IPv6 address.

configutil/normalize.go
```diff
@@ -0,0 +1,106 @@
...
+			// this is a unique case, host is ipv6 and requires brackets due to
+			// being part of a url, but they won't be added by net.JoinHostPort
+			// as there is no port
```
jefferai, line 25: Suggest putting the relevant RFC link here describing how this URL handling differs from other ways you format IPv6

configutil/normalize.go
```diff
@@ -0,0 +1,106 @@
...
+	if address == "" {
+		return "", fmt.Errorf("empty address")
+	}
+
+	if ip := net.ParseIP(address); ip != nil {
+		return ip.String(), nil
+	}
+
+	if strings.HasPrefix(address, "[") && strings.HasSuffix(address, "]") {
+		return NormalizeAddr(address[1 : len(address)-1])
```
jefferai, line 74: Love recursion. Also love recursion bombing by sending in a user-defined address with 32k `[]` pairs around an address.

Safer to just put that into a for loop until that condition doesn't hold true anymore, then run the empty and net.ParseIP checks again.

kheina: oh geeze, good point

jefferai (dismissed):

louisruch (approved):
Great work on this thanks!

hugoghx (dismissed):
🚢

jefferai (approved):

PR #160, opened 2025-03-19
Make Web ID Providers and Environment/Shared Credential Providers mutually exclusive

vinay-gopalan:
HVD uses an AWS Profile in its environment, and a few customers use environment variables for other AWS operations like KMS while attempting to use WIF. In the current implementation all types of credentials are appended to the credential chain if they are detected, and using environment variables for 1 account while attempting to assume a role using web identity for another account leads to conflicting errors and causes WIF to be unusable in such cases.

This PR only enables environment and shared credential providers if web identity providers are not explicitly enabled. If no WIF parameters were explicitly set, AWS Profile or environment variables will work as they do today.

### DIFF
--- a/awsutil/generate_credentials.go
+++ b/awsutil/generate_credentials.go
@@ -205,28 +205,6 @@ func (c *CredentialsConfig) GenerateCredentialChain(opt ...Option) (*credentials
 		c.log(hclog.Debug, "added static credential provider", "AccessKey", c.AccessKey)
 	}
 
-	// Add the environment credential provider
-	if opts.withEnvironmentCredentials {
-		providers = append(providers, &credentials.EnvProvider{})
-		c.log(hclog.Debug, "added environment variable credential provider")
-	}
-
-	// Add the shared credentials provider
-	if opts.withSharedCredentials {
-		profile := os.Getenv("AWS_PROFILE")
-		if profile != "" {
-			c.Profile = profile
-		}
-		if c.Profile == "" {
-			c.Profile = "default"
-		}
-		providers = append(providers, &credentials.SharedCredentialsProvider{
-			Filename: c.Filename,
-			Profile:  c.Profile,
-		})
-		c.log(hclog.Debug, "added shared credential provider")
-	}
-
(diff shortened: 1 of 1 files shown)


awsutil/generate_credentials.go
```diff
@@ -240,6 +218,8 @@ func (c *CredentialsConfig) GenerateCredentialChain(opt ...Option) (*credentials
 	if roleSessionName == "" {
 		roleSessionName = os.Getenv("AWS_ROLE_SESSION_NAME")
 	}
+
+	var webIDProviderEnabled bool
```
benashz, line 222: I wonder if you had considered adding a new environment variable that controls whether or not creds should be added from the standard AWS env vars? The proposed approach to tie that to the web identity token feels like a bit of a surprise. If the two credential sources are mutually exclusive, perhaps an error should be raised if both are set at the same time?

I'd be happy to discuss further offline.

vinay-gopalan: With HVD's design, it looks like there are always two credential sources being defined when using AWS Secrets WIF in Vault. HVD due to its environment configuration uses an AWS Profile, and since Vault is running in the same environment, AWS Secrets Engine WIF running on HVD also picks up this profile, leading to two credential sources (an AWS Profile in the environment and WIF parameters). At present, WIF is unusable because of this.

I don't think returning an error when two credential sources are set would fix the issue, since WIF on HVD will continue to always error out (as it does today). 

I thought about introducing an environment variable to select a credential type, but since the environment variables will also be shared between HVD and the AWS Secrets Engine, I think this will also lead us to a similar issue — AWS Secrets Engine will still detect the new `USE_ENVIRONMENT_CREDENTIAL_PROVIDER` (placeholder name), and attempt to read credentials from the environment, thereby bypassing and not using WIF

Lmk if that makes sense! Will sync up offline as well

vinay-gopalan:
After syncing up on this, we decided it is better to disable these on the AWS Secrets Engine side instead since the functionality for that exists through providing `Options` when generating credential chains

PR #192, opened 2026-09-08
fix(awsutil): Fix role arn fall-through bug

delainerogers:
https://hashicorp.atlassian.net/browse/ICU-19916
When RoleARN is set but no `~/.aws/config` file contains a matching `role_arn` entry, the previous code used `config.WithAssumeRoleCredentialOptions`, a modifier that is only invoked if the AWS SDK independently decides to create an `AssumeRoleProvider`, which it does not do without that config file entry. This meant the modifier was silently ignored and all API calls ran as the ambient identity instead of the intended role, with no error. The fix constructs an `stscreds.AssumeRoleProvider` explicitly after `LoadDefaultConfig` and assigns it directly to `awsConfig.Credentials`, so `sts:AssumeRole` is always called when RoleARN is set programmatically regardless of what is in `~/.aws/config`.

[Doc showing the before/after output](https://ibm-my.sharepoint.com/:w:/r/personal/matt_romanelli_ibm_com/Documents/ICU-065/Bug%20fix%20terminal%20output.docx?d=w6566a9601ac946c3829fe087c7522ff9&csf=1&web=1&e=cu37dk)

## PCI review checklist

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've worked with GRC to document the impact of any changes to security controls.

  Examples of changes to controls include access controls, encryption, logging, etc.

- [ ] If applicable, I've worked with GRC to ensure compliance due to a significant change to the in-scope PCI environment.

  Examples include changes to operating systems, ports, protocols, services, cryptography-related components, PII processing code, etc.


awsutil/generate_credentials.go
```diff
@@ -247,24 +247,9 @@ func (c *CredentialsConfig) generateAwsConfigOptions(ctx context.Context, opts o
...
-						Key:   aws.String(k),
-						Value: aws.String(v),
-					})
-				}
-			})
-			cfgOpts = append(cfgOpts, assumeRoleCred)
-			c.log(hclog.Debug, "added ec2-instance role provider", "roleARN", c.RoleARN)
 		}
+		// The plain assume-role case (no web-identity token) is handled
+		// directly in GenerateCredentialChain.
```
ryanflynn9299, line 252: This is a good pointer to the longer comment, but could we add some commentary on what the consequences of having the deleted block here are, so future contributors are weary of the circumstances that precipitated this bug in the first place? (E.g. processing the role_arn here resulted in lazy AssumeRoleProvider instantiation... etc)

matt-romanelli:
A few notes on the PR description:
- I like that you added the Jira link
- It can be beneficial to add a brief explanation on what the problem is (without getting into the technical details)
- This could be a Matt problem but I had some difficulty understanding what was said in the description (it's pretty dense)
- I'd like to know how this was tested. Were you able to reproduce the issue before and then you validated that it didn't exist afterwards? What test cases did you walk through?

ryanflynn9299 (dismissed):
Looks good!

awsutil/generate_credentials.go
```diff
@@ -292,6 +277,38 @@ func (c *CredentialsConfig) GenerateCredentialChain(ctx context.Context, opt ...
...
+		if opts.withSTSAPIFunc != nil {
+			stsClient, err = opts.withSTSAPIFunc(&awsConfig)
+			if err != nil {
+				return nil, fmt.Errorf("error creating STS client: %w", err)
+			}
+		} else {
+			stsClient = sts.NewFromConfig(awsConfig)
+		}
+		provider := stscreds.NewAssumeRoleProvider(stsClient, c.RoleARN, func(o *stscreds.AssumeRoleOptions) {
+			if c.RoleSessionName != "" {
```
matt-romanelli, line 295: Are we certain that RoleSessionName and RoleExternalId would be "" and not nil?

awsutil/generate_credentials.go
```diff
@@ -292,6 +277,38 @@ func (c *CredentialsConfig) GenerateCredentialChain(ctx context.Context, opt ...
...
+			}
+			for k, v := range c.RoleTags {
+				o.Tags = append(o.Tags, types.Tag{
+					Key:   aws.String(k),
+					Value: aws.String(v),
+				})
+			}
+		})
+		awsConfig.Credentials = aws.NewCredentialsCache(provider)
+		c.log(hclog.Debug, "added explicit assume-role provider", "roleARN", c.RoleARN)
```
matt-romanelli, line 309: I see that this is added as just a debug log. I feel like it could be beneficial to add some logs throughout this logic that are not just debug - seeing as Boundary admins use the logs that are outputted.

awsutil/generate_credentials.go
matt-romanelli: I am also don't think we should have comments floating at the bottom of a function. It makes sense right now because we know that previous logic was removed. But in the future, people won't know that and will just see a comment essentially saying "we're not doing something here".

With that being said, could you explain the differences between `generateAwsConfigOptions` and `GenerateCredentialChain`? Why should we move the logic?

matt-romanelli (commented):
Overall, the code looks good and clean. I just left some comments looking for some further clarification.

awsutil/generate_credentials.go
ryanflynn9299: Yes [the zero value of a string](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://go.dev/tour/basics/12&ved=2ahUKEwi9kK-1guKWAxV4M1kFHfumIdMQFnoECBwQAQ&usg=AOvVaw06NERTG6bE8_n-X71QxH5K) is `""`, while the [zero value of a string pointer](https://go.dev/tour/moretypes/1) is nil. Since RoleSessionName, RoleExternalId and RoleArn are type `string`, their zero values will be `""`

matt-romanelli: Got it - sounds good. Thanks for the clarification!

awsutil/generate_credentials.go
delainerogers: Will remove the comment, I agree that it's messy to keep it for future readers. I'll try my best to explain what's going on here but the aws flow is pretty convoluted 
`generateAwsConfigOptions` is basically filling out a form (the config options)
`GenerateCredentialChain` takes the filled-out form (config options) and uses it to actually set up `aws.Config`
Once we have an `aws.Config`, we can create an STS client. Using the STS client, we can create an `AssumeRoleProvider`.
Explicitly setting up the `AssumeRoleProvider` fixes the bug; before, the `awsConfig` wouldn't actually contain an `AssumeRoleProvider`, so when we wanted to assume role, there was no provider to do it, so it was skipped
So the `awsConfig` needs to contain the `AssumeRoleProvider`, but we need an existing `awsConfig` to even create the `AssumeRoleProvider`. Therefore we moved the logic to after the config is created
Does this make sense?

awsutil/generate_credentials.go
delainerogers: updated the logs

awsutil/generate_credentials.go
```diff
@@ -292,7 +275,42 @@ func (c *CredentialsConfig) GenerateCredentialChain(ctx context.Context, opt ...
...
+			if err != nil {
+				return nil, fmt.Errorf("error creating STS client: %w", err)
+			}
+		} else {
+			stsClient = sts.NewFromConfig(awsConfig)
+		}
+		provider := stscreds.NewAssumeRoleProvider(stsClient, c.RoleARN, func(o *stscreds.AssumeRoleOptions) {
+			if c.RoleSessionName != "" {
+				o.RoleSessionName = c.RoleSessionName
+			}
```
matt-romanelli, line 299: Referencing Ryan's PR for getting the target ec2client, it looks like we may want to provide a default session name? @ryanflynn9299 

https://github.com/hashicorp/boundary-plugin-aws-internal/pull/32/changes#diff-776632a5e72c8ff33a93210cd14beacd88366b9cf94af41e673e4a94b7f5484cR772

awsutil/generate_credentials.go
matt-romanelli: Yep! Thanks!

awsutil/generate_credentials.go
delainerogers: From what I can tell, the AWS SDK generates a session name if one doesn't exist every time the credentials are retrieved, so there shouldn't be any issues with STS calls with empty session names. However, it might be cleaner to follow what @ryanflynn9299 did and set the session name to a constant instead of a new generated name each time if that's a concern
```
// Retrieve generates a new set of temporary credentials using STS.
func (p *AssumeRoleProvider) Retrieve(ctx context.Context) (aws.Credentials, error) {
	// Apply defaults where parameters are not set.
	if len(p.options.RoleSessionName) == 0 {
		// Try to work out a role name that will hopefully end up unique.
		p.options.RoleSessionName = fmt.Sprintf("aws-go-sdk-%d", time.Now().UTC().UnixNano())
	}
```

awsutil/generate_credentials.go
```diff
@@ -292,7 +275,42 @@ func (c *CredentialsConfig) GenerateCredentialChain(ctx context.Context, opt ...
...
+			for k, v := range c.RoleTags {
+				o.Tags = append(o.Tags, types.Tag{
+					Key:   aws.String(k),
+					Value: aws.String(v),
+				})
+			}
+		})
+		awsConfig.Credentials = aws.NewCredentialsCache(provider)
+		c.log(hclog.Info, "configured assume-role credential provider", "roleARN", c.RoleARN)
```
louisruch: Should this be debug?

awsutil/generate_credentials.go
```diff
@@ -292,7 +275,42 @@ func (c *CredentialsConfig) GenerateCredentialChain(ctx context.Context, opt ...
...
+	// role_arn entry.
+	if c.RoleARN != "" && c.WebIdentityTokenFile == "" && c.WebIdentityToken == "" {
+		var stsClient stscreds.AssumeRoleAPIClient
+		if opts.withSTSAPIFunc != nil {
+			stsClient, err = opts.withSTSAPIFunc(&awsConfig)
+			if err != nil {
+				return nil, fmt.Errorf("error creating STS client: %w", err)
+			}
+		} else {
+			stsClient = sts.NewFromConfig(awsConfig)
```
louisruch: So there is a `c.STSClient` the seems to most of what you are doing here but more, it also takes the options and uses them - I am specifically curious about `WithEndpointResolverV2` that we are potentially dropping here

louisruch (commented):
Thanks for digging into this left one comment where I am concerned we might be dropping STSEndpointResolver

awsutil/generate_credentials.go
ddebko: Yea that is a really good catch, we need the custom endpoint resolver so that we can enable dualstack(IPv4 & IPv6) in our storage and host plugin services.

delainerogers: Thanks for catching this! Will fix

awsutil/generate_credentials.go
ryanflynn9299: FYI, more information to help with the resolution of this: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_iam-condition-keys.html#ck_rolesessionname

also: https://docs.aws.amazon.com/sdkref/latest/guide/feature-assume-role-credentials.html

ddebko (approved):
I'm happy with the fix for the dualstack issue, feel free to merge once all the other comments are resolved.

louisruch (approved):

awsutil/generate_credentials.go
delainerogers: Based on the slack conversation, we're going to keep it as is

